PR #21, opened 2015-11-16
Coveralls Coverage

rasbt:
Adding coverage reports and badge via coveralls.io -- also here, you need to log in and enable the repo.
Plus, I added Python 2.7 and 3.5 version badges and a GPLv3 license badge.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,3 @@
+branch = True
+source = tpot
+include = */tpot/*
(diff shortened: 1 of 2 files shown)


rhiever:
Hmmm, there's no free version for coveralls.io?

rasbt:
Coveralls should be free; I am using it for free :)

rasbt:
See here: https://coveralls.io/pricing 

> OPEN 
SOURCE
Coveralls is and will always be free for open source projects. You can always add a subscription to get coverage reports on private repos later.

rhiever:
Gotcha. I think I was confused by their aggressive push on upgrading to the pro version. I think it's set up now.

PR #22, opened 2015-11-16
code cleanup

rasbt:
Cleaning up some style issues.

Question: I didn't want to make any changes to the docstrings because this is a personal taste thing, and I am not sure what you think about it, but what about using standard Python docstrings? Also, the docstrings are not very informative, e.g., if someone uses `help(TPOT)`, there wouldn't be much information on the attributes and methods. I'd recommend to change that, also with respect to a future auto-generated API documentation.

### DIFF
--- a/tpot/tpot.py
+++ b/tpot/tpot.py
@@ -38,8 +38,8 @@ from deap import creator
 from deap import tools
 from deap import gp
 
-class TPOT:
-    
+class TPOT(object):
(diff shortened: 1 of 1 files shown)


rhiever:
Thanks for all these PRs. This is great.

Yes, I think we should improve the docstrings. Please file an issue for that so we don't forget. :-)

PR #23, opened 2015-11-16
Docstrings & Private Methods

rasbt:
I added the docstrings to the public methods, please double-check if this is correct in its meaning :).
Also, I made the internal methods private so that a user is not confused when he/she calls help(TPOT)

### DIFF
--- a/tests.py
+++ b/tests.py
@@ -16,7 +16,8 @@ from sklearn.tree import DecisionTreeClassifier
 
 # Set up the iris data set for testing
 mnist_data = load_digits()
-training_features, testing_features, training_classes, testing_classes = train_test_split(mnist_data.data, mnist_data.target, random_state=42)
+training_features, testing_features, training_classes, testing_classes =\
+        train_test_split(mnist_data.data, mnist_data.target, random_state=42)
 
 training_data = pd.DataFrame(training_features)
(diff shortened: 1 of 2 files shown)


rasbt:
Hah, weird, somehow my local nosetest didn't complain ... have to look into this ...

rhiever:
I imagine the tests are failing because those private methods are no longer available for inspection.

rasbt:
> I imagine the tests are failing because those private methods are no longer available for inspection.

Yes, and I was running nosetests in the tpot/tpot dir so that it actually didn't run the tests.py at all :P

rhiever:
Ah yes, that would do it. :-)

I'm wondering if the tests should be moved to a separate directory.

JanSchulz:
@rasbt Are these all completely handcrafted docstrings or do you have a tool for that?

rasbt:
@JanSchulz 
These were all "manually" written, I don't think this can be done via a tool (since it need to guess your intention as programmer). But maybe we can automate the type-checking part some day ;) (https://www.python.org/dev/peps/pep-0484/)

JanSchulz:
@rasbt Ok, thought you have a tool which at least generate some of the stuff... I currently use pycharm at it at least generates  the structure and the names of the variables...

Issue #36, opened 2015-12-01
Output the current best pipeline when the user terminates TPOT on the command line


rhiever:
Issue raised from #34

Issue #5, opened 2015-11-11
Output the pipeline to sklearn code


rhiever:
Provide a utility function to export the pipeline as sklearn code.

PR #37, opened 2015-12-02
First pass at pipeline export functionality

rhiever:
Per #5

This PR adds the `export()` function, which allows the user to specify an output file for the pipeline. Once `export()` is called (and a pipeline has already been optimized), this function converts the pipeline into its corresponding Python code and exports it to the specified output file.

Docs have been updated along with this PR to demonstrate how the `export()` function works.

@rasbt, I would greatly appreciate if you could review this before we merge it.


rhiever:
The decline in coverage is expected. The new `export()` function introduces a fair bit of new code without a unit test for it (yet). I'll eventually get to expanding the unit tests... :-)

tpot/tpot.py
```diff
@@ -288,6 +289,182 @@ def score(self, training_features, training_classes, testing_features, testing_c
...
+        Parameters
+        ----------
+        output_file_name: string
+            String containing the path and file name of the desired output file
+
+        Returns
+        -------
+        None
+        """
+        if self.optimized_pipeline_ == None:
```
rasbt: if self.optimized_pipeline_ == None:

should be

    if not self.optimized_pipeline_:

rasbt: This should raise a `ValueError`, right?

tpot/tpot.py
```diff
@@ -288,6 +289,182 @@ def score(self, training_features, training_classes, testing_features, testing_c
...
+            raise Exception('A pipeline has not yet been optimized. Please call fit() first.')
```
rasbt: This should raise a ValueError, right?

tpot/tpot.py
```diff
@@ -288,6 +289,182 @@ def score(self, training_features, training_classes, testing_features, testing_c
...
+        
+        exported_pipeline = self.optimized_pipeline_
+        
+        # Replace all of the mathematical operators with their results
+        while True:
```
rasbt: I am not sure if I understand this structure correctly. Why do we need an outer 

    while True:
        ...
        else:
            break

here? Can't this simply be a for loop?

tpot/tpot.py
```diff
@@ -288,6 +289,182 @@ def score(self, training_features, training_classes, testing_features, testing_c
...
+            else:
+                break
+
+        # Unroll the nested function calls into serial code
+        pipeline_list = []
+        result_num = 1
+        while True:
+            for node_index in range(len(exported_pipeline) - 1, -1, -1):
+                node = exported_pipeline[node_index]
+                if type(node) != deap.gp.Primitive:
```
rasbt: Can it be written as `isinstance(node, deap.gp.Primitive)` ?

tpot/tpot.py
```diff
@@ -288,6 +289,182 @@ def score(self, training_features, training_classes, testing_features, testing_c
...
+                    exported_pipeline = exported_pipeline[:i] + [new_val] + exported_pipeline[i + 3:]
+                    break
+            else:
+                break
+
+        # Unroll the nested function calls into serial code
+        pipeline_list = []
+        result_num = 1
+        while True:
+            for node_index in range(len(exported_pipeline) - 1, -1, -1):
```
rasbt: General question: Is there a reason why you are looping backwards? Can't this be replaced by 

    for node_index in enumerate(exported_pipeline):

?

rasbt:
> @rasbt, I would greatly appreciate if you could review this before we merge it.

Sure, I can take a more detailed look at it later when I get home. However, would it be possible to add a small unit test comparing the actual output with what you'd expect? I think that would be helpful as "documentation" and to follow along.

tpot/tpot.py
rhiever: Isn't the equality more explicit?

I just noticed in the `score()` function it now says:

    if self.optimized_pipeline_ is None:

Personally, I think the equality is better since it is explicitly checking for the default state (`None`). I'm open to being convinced otherwise though.

tpot/tpot.py
rhiever: I struggled with the proper exception for a bit because the optimized pipeline is technically not a value passed to a parameter. I suppose ValueError is the best fit here if we didn't want to raise a generic Exception.

tpot/tpot.py
rhiever: I *think* you're right that it's unnecessary to repeatedly start over when looping through the TPOT stack, but it is necessary to loop through the function tree backwards (which will answer your question below as well).

The TPOT stack is stored as a list of functions and primitives, so for example,

`random_forest(decision_tree(dataset, 1+1, 2+2), 3+3, 4+(4+4)` would be stored as:

`[random_forest, decision_tree, dataset, +, 1, 1, +, 2, 2, +, 3, 3, +, +, 4, 4, +, 4, 4]`

Thus you can't run through the stack forwards because the parameters to the first functions have not been completely evaluated yet.

Therefore to deconstruct the TPOT stack, you need to proceed backwards through the stack and evaluate the functions as you run into them.

So proceeding backwards, you would first run into the `+`. You know that `+` accepts two parameters, so you take the two parameters that follow -- `4` and `4`. Evaluate `4+4` and replace those 3 entries by the result, `8`. Now you have:

`[random_forest, decision_tree, dataset, +, 1, 1, +, 2, 2, +, 3, 3, +, +, 4, 4, 8]`

You again run into a `+`, the following two parameters are `4` and `4`, repeat above...

`[random_forest, decision_tree, dataset, +, 1, 1, +, 2, 2, +, 3, 3, +, 8, 8]`

... and I think you get it from there. That leaves the following stack once all the mathematical operators are replaced:

`[random_forest, decision_tree, dataset, 2, 4, 6, 16]`

Next you run into `decision_tree`. You know that it takes 3 parameters, so you evaluate `decision_tree(dataset, 2, 4)` and replace its entry with the result (say, `result1`):

`[random_forest, result1, 6, 16]`

Then you finally have `random_forest`, which similarly takes 3 parameters, and you can evaluate the last call of the stack: `random_forest(result1, 6, 16)`

Make sense?

tpot/tpot.py
rhiever: What's the advantage of `isinstance()`?

tpot/tpot.py
rhiever: See https://github.com/rhiever/tpot/pull/37#discussion_r46484892

rhiever:
Sure, here's an example:

    from tpot import TPOT
    from sklearn.cross_validation import train_test_split

    digits = load_digits()
                                                        train_size=0.75)

    tpot = TPOT(generations=1, verbosity=2)
    tpot.fit(X_train, y_train)
    print(tpot.score(X_train, y_train, X_test, y_test))
    tpot.export('tpot_export.py')

TPOT will likely discover that a random forest alone does well on the data set, so `tpot_export.py` should contain something like:

    from itertools import combinations

    import numpy as np
    import pandas as pd

    from sklearn.cross_validation import StratifiedShuffleSplit
    from sklearn.tree import DecisionTreeClassifier
    from sklearn.ensemble import RandomForestClassifier

    # NOTE: Make sure that the class is labeled 'class' in the data file
    tpot_data = pd.read_csv('PATH/TO/DATA/FILE', sep='COLUMN_SEPARATOR')
    training_indeces, testing_indeces = next(iter(StratifiedShuffleSplit(tpot_data['class'].values, n_iter=1, train_size=0.75)))


    result1 = tpot_data.copy()

    # Perform classification with a random forest classifier
    rfc1 = RandomForestClassifier(n_estimators=97, max_features=min(8, len(result1.columns) - 1))
    rfc1.fit(result1.loc[training_indeces].drop('class', axis=1).values, result1.loc[training_indeces]['class'].values)
    result1['rfc1-classification'] = rfc1.predict(result1.drop('class', axis=1).values)

If you want to perform further inspection on the pipelines, if you `print()` the optimized pipeline at any point it will give you the nested function version, from which you can deduce what the linear code version should look like.

rhiever:
Oh, let's make sure to close #36 when this is merged. Now when the user ends TPOT early on the command line and they provided an output file, the current best pipeline will be exported.

tpot/tpot.py
rasbt: Oh I see;  I thought this way it may be more robust, accounting for empty objects like "", [], {}. But maybe being explicit is not a bad idea here.

 However

    if self.optimized_pipeline_ is None:

is definitely preferred over 

    if self.optimized_pipeline_ == None:

It's more efficient (because you don't have the `__eq__` overhead) and also cleaner style.

tpot/tpot.py
rasbt: I think it's not that important but may be useful for a user and the unit tests in certain cases.

rasbt:
> Oh, and let's make sure to close #36 when this is merged. Now when the user ends TPOT early on the command line and they provided an output file, the current best pipeline will be exported.

Sounds cool! Maybe a nice addition would be to refactor the `export`method a bit so that a slimmer "core" component can be written to a log-file. This let's the user also conveniently check the progress over time (e.g., via `tail tpot_run_x.log`) or so.

tpot/tpot.py
rasbt: It would support inheritance, but I am not sure if this will be useful/beneficial down the road. A small cosmetic change though:

    if type(node) is not deap.gp.Primitive: 

;)

tpot/tpot.py
rasbt: Thanks for the detailed description, that's really helpful!

rhiever:
`export()` actually runs pretty quick even on large pipelines, so it wouldn't be much overhead to run it every generation.

tpot/tpot.py
rhiever: Looking at the while loop again, I think I have it there because the code is constantly changing the contents of the list it's iterating over, so it's safer to restart the iterator over again when the list is changed. That probably doesn't matter since we're iterating over the list backwards, but I'll leave that for a future revision of the export method (since it's pretty fast right now anyway).

rasbt:
> export() actually runs pretty quick even on large pipelines, so it wouldn't be much overhead to run it every generation.

Oh, sure, but I was more thinking of "bloated log files" here. I think that log/status files are generally helpful to keep track of errors but also to judge the progress if your are running stuff remotely, and analyzing what's going on under the hood (e.g., think bakc of ye goode olde times submitting jobs to HPCC ;)). So, I was thinking to refactor it into a more bare-bones "export_params" and a "export_pipeline_standalone" or so. But this is just a general suggestion, doesn't have to be now :)

rasbt:
Besides that, the code looks fine to me so far. But I haven't run it through a debugger yet and looked at it in detail... sorry, it's the end of the year and I am pretty busy wrapping things up before I go on a family visit, but January should be a good time for a fresh start :)

rhiever:
Alrighty, I think I'll merge this for now since I'd like to push this functionality out. Please file bug reports against it if you see anything wrong.

Issue #33, opened 2015-11-30
Performance Metrics & Fitness Functions


bartleyn:
I imagine that it would be useful to use TPOT to find models that optimize alternate performance metrics like precision, recall, etc. As such, I've come up with the following brainstorming questions:

1. Does having alternate performance metrics/fitness functions make sense for the users?
2. Does it make sense to add alternate metrics when reporting the best model? If so, which metric do we use as the fitness function?
3. Since there is native support for multi-class/label classification regular precision, recall, and F1 may not be that useful. Should we just take the average version of scores like these when necessary?

There are plenty of other questions, but I figured this would be a decent place to start. Let me know if I'm totally misguided in proposing this -- I won't profess to be an expert in genetic programming.

rhiever:
>Does having alternate performance metrics/fitness functions make sense for the users?

Yes. I think we should eventually support allowing the user to pass arbitrary scoring functions, similar to how sklearn does it.

>Does it make sense to add alternate metrics when reporting the best model? If so, which metric do we use as the fitness function?

As in, use one scoring functions for optimization then a different scoring metric for final model selection? Interesting idea. I'm currently working on a version of TPOT that allows it to optimize on multiple criteria simultaneously, so perhaps that will help in this regard.

>Since there is native support for multi-class/label classification regular precision, recall, and F1 may not be that useful. Should we just take the average version of scores like these when necessary?

That's what we currently do with accuracy. I think it makes sense to do the same with other measures.

bartleyn:
>Yes. I think we should eventually support allowing the user to pass arbitrary scoring functions, similar to how sklearn does it.

Would passing in a keyword for the specific metric work for now?

> As in, use one scoring functions for optimization then a different scoring metric for final model selection? Interesting idea. I'm currently working on a version of TPOT that allows it to optimize on multiple criteria simultaneously, so perhaps that will help in this regard.

Yeah, I was thinking that and/or adding additional metrics to the report of the final model. 

>That's what we currently do with accuracy. I think it makes sense to do the same with other measures.

I'm happy to contribute by adding simple support for F1/Precision/Recall in the same vein as the accuracy. I think it should be relatively straightforward.

rhiever:
>>Yes. I think we should eventually support allowing the user to pass arbitrary scoring functions, similar to how sklearn does it.

>Would passing in a keyword for the specific metric work for now?

[...]

PR #38, opened 2015-12-03
Arbitrary scoring functions + doc example

bartleyn:
Per #33 , I've taken a stab at the arbitrary scoring function implementation, and wrote an example using scoring functions written in the same vein as the original balanced accuracy calculation.

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -20,6 +20,7 @@ pages:
   - examples/Using_TPOT_via_the_command_line.md
   - examples/Using_TPOT_via_code.md
   - examples/MNIST_Example.md
+  - examples/Custom_Scoring_Functions.md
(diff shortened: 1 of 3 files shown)


rhiever:
Looks good to me. Thoughts @rasbt?

rhiever:
Merging this now. Thank you @bartleyn!

bartleyn:
I'm happy to take another look and see if there's a smarter/more efficient way to handle this -- I imagine passing the dataframes to the scoring function will get bulky after a point.

rhiever:
We're definitely open to suggestions. :-) It's nice having DataFrames because you can (theoretically) maintain the names of the columns through the whole process, which helps with inspection if you're, e.g., looking at feature importance scores.

Issue #39, opened 2015-12-03
Implement more classifier pipeline operators


rhiever:
Similar to the Decision Tree and Random Forest classifier pipeline operators, also implement:

* [x] Logistic regression: http://scikit-learn.org/stable/modules/linear_model.html#logistic-regression

* [x] SVM: http://scikit-learn.org/stable/modules/svm.html#classification

* [x] kNN: http://scikit-learn.org/stable/modules/neighbors.html#nearest-neighbors-classification

* [x] Gradient boosting: http://scikit-learn.org/stable/modules/generated/sklearn.ensemble.GradientBoostingClassifier.html

@rasbt, do you think we should add any more than this? I'd like to add ANNs eventually, but since they're not directly supported in sklearn, that will wait for a later time.

rhiever:
Just to bug you even more, @rasbt, check out https://github.com/rhiever/tpot/commit/96c69a4fd260e1dfeeed1e360943a86803204cb7 for the implementation of these new classifiers.

Any other parameters that you would include?

Would you implement any parameters differently?

rhiever:
@amueller, do you have any advice on what additional model parameters to open up to evolution?

Chris7:
Nice to see classifiers are coming along!

One thing I've noticed is you have a ton of repetitive code and it seems like you could abstract the implementation of all these methods to just a generic_regressor/generic_classifier function and a common structure for storing the regressors/parameters. Have you tried any generic approach yet?

amueller:
ANNs are in dev. They don't have dropout yet, but will soon.

How about gradient boosting?

rhiever:
@Chris7: I noticed that today too when I was implementing the new classifiers. I'll look into abstracting the common bits next week, as that would indeed save quite a lot of repeated code.

@amueller: :+1: Looking forward to ANNs in sklearn. I'll add gradient boosting as well. Do you have a sense for what are the 2-3 most important parameters (if there are that many) for each model? I've tinkered around with various parameters for various models but don't have a comprehensive view of them like you might.

amueller:
We want to add a "Default grid" https://github.com/scikit-learn/scikit-learn/pull/5564 but it is somewhat stalled. I'm crazy busy at the moment but I hope to work on that soonish.

rhiever:
Nice. Looks like [this](https://github.com/zermelozf/scikit-learn/blob/default-cv/sklearn/grid_search_default.py) will be a good start. Thank you @amueller!

```python
[...]

PR #40, opened 2015-12-04
Add more classifiers per #39

rhiever:


bartleyn:
> training_features = input_df.loc[input_df['group'] == 'training'].drop(['class', 'group', 'guess'], axis=1).values
I've always wondered -- is it faster to drop columns or to create a slice all the other columns?

rhiever:
I've never checked to be honest. My assumption has always been that dropping a few specific columns is faster than finding out the column names of all the *other* columns and slicing down to those.

bartleyn:
I agree with your intuition -- out of curiosity I'll take some time later tonight and see if there are any existing empirical test results pointing one way or the other, otherwise I'll run my own tests and see.

PR #53, opened 2015-12-08
fixing column masking & typo

rcarneva:
Two find-and-replaces:

1. Column masking in the exported pipeline wasn't working correctly for me with SelectKBest on a model since the mask was coming through as a numeric but some of the columns were named. Updated `[mask]` -> `.iloc[:, mask]` and it seems to work now, though more testing would be good if it was working as expected on your end.
2. `indeces` -> `indices`

### DIFF
--- a/tpot/tpot.py
+++ b/tpot/tpot.py
@@ -213,16 +213,16 @@ class TPOT(object):
             np.random.shuffle(data_columns)
             training_testing_data = training_testing_data[data_columns]
 
-            training_indeces, testing_indeces = next(iter(StratifiedShuffleSplit(training_testing_data['class'].values,
+            training_indices, testing_indices = next(iter(StratifiedShuffleSplit(training_testing_data['class'].values,
                                                                                  n_iter=1,
                                                                                  train_size=0.75,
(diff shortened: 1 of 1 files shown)


rhiever:
Thanks @rcarneva! Both versions work fine on my end, though I believe the `.iloc[:, mask]`syntax is "more correct." What version of Python/pandas are you running?

rcarneva:
Python 2.7.10 / pandas 0.17.1

rhiever:
Interesting. I'm on Python 3.5 / pandas 0.17.1 and both cases work.

rcarneva:
Think I tracked it down. The two features in my training set had (integer) names 1 and 2, and so pandas seemed to have trouble applying the integer mask. Renaming the columns to strings made it work with the original code, but the updated version with `iloc` seems to work in both situations, so is probably a safer bet.

BTW, really neat idea and project.

rhiever:
I had a feeling that was it. Thanks for looking into it!

>BTW, really neat idea and project.

Thanks! Happy to have you involved more if you have any ideas or any of the open issues catch your fancy.

PR #54, opened 2015-12-08
swapping f_classif for chi2 to allow negative features

rcarneva:
Another small change -- swapped in `f_classif` for the feature selection processes since `chi2` gets tripped up on negative values in the design matrix.


rhiever:
Funny -- I just did this too when I was testing the feature preprocessing operators I implemented. Cheers!

Issue #43, opened 2015-12-05
Create an abstract function for the classifier pipeline operators


rhiever:
There's quite a bit of repeated code in the classifier functions. Abstract the common code to a single function then have the classifiers call that function and pass their custom parameters/model.

bartleyn:
I've got a long plane ride tomorrow evening that I could use to knock this out, especially now that I'm more familiar with DEAP.

rhiever:
Awesome!

On Wednesday, December 9, 2015, Nathan <notifications@github.com> wrote:

> I've got a long plane ride tomorrow evening that I could use to knock this
> out, especially now that I'm more familiar with DEAP.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/rhiever/tpot/issues/43#issuecomment-163486702>.
>


-- 
Randal S. Olson, Ph.D.
Postdoctoral Researcher, Institute for Biomedical Informatics
University of Pennsylvania

E-mail: rso@randalolson.com | Twitter: @randal_olson
<https://twitter.com/randal_olson>
http://www.randalolson.com

dmarx:
I think a better approach would be to create a class for each operator, where the separate operators would inherit from base classes that define the expected API similar to how scikit-learn is organized. If done correctly, this would greatly simplify adding any arbitrary operator in the future, including empowering users to create customized operators. I'm thinking of a framework that would look something like this (non-working demo to roughly demonstrate the idea): https://gist.github.com/dmarx/7d72263a02b82cd276e1

bartleyn:
Agreed with the oo approach, as among other things it'll also avoid bloating the main TPOT class down the line. Just to flesh this idea out, where would you propose handling input validation? An abstract function that each operator implements?

dmarx:
We could probably attach that to the base class. In the gist, I included `inputtypes` as an input parameter to the base class with the intention of using that to pass the appropriate values to `pset.addPrimitive`. I imagine we could add a simple input validation function to the base class as well that would access these values.

Personally, I'm more concerned about what the best way would be to tie instances of these "operator" objects with `deap` pipeline nodes. I'm new to deap (and only just started looking at `tpot` today as well) so that might actually be a fairly trivial consideration.

Do we even need a separate input validation function each class? Or does `deap.gp.PrimitiveSetTyped.addPrimitive` already handle input validation for us?

bartleyn:
[...]

PR #57, opened 2015-12-14
Abstract function for the classifier pipeline operators

bartleyn:
Per #43 , this abstracts the shared code between static models into a static method that each model can call. In addition, I built a test for testing all the static models / this new abstract method indirectly, and tweaked the num_trees parameter in random_forest to n_estimators to facilitate testing. 

Potential improvements:
1) Maybe it's just me, but It feels a little odd to call a static function from another static function. Does it make sense to move this method (and perhaps some others) to a utility function module? Either way, is it faster to call this abstract method as a static method inside the class, a class method inside the class, or something else?
2) Is there a more appropriate name for the method?
3) Is there a better place to do the input validation (i.e. if there are only three columns in the input_df)  than within this nested function call?

### DIFF
--- a/tpot/tpot.py
+++ b/tpot/tpot.py
@@ -724,31 +724,7 @@ else:
         if max_depth < 1:
             max_depth = None
 
-        # If there are no features left (i.e., only 'class', 'group', and 'guess' remain in the DF), then there is nothing to do
-        if len(input_df.columns) == 3:
-            return input_df
-        
-        input_df = input_df.copy()
-
-        training_features = input_df.loc[input_df['group'] == 'training'].drop(['class', 'group', 'guess'], axis=1).values
-        training_classes = input_df.loc[input_df['group'] == 'training', 'class'].values
-
-        dtc = DecisionTreeClassifier(max_features=max_features,
-                                     max_depth=max_depth,
-                                     random_state=42)
-
-        dtc.fit(training_features, training_classes)
-
-        all_features = input_df.drop(['class', 'group', 'guess'], axis=1).values
-        input_df.loc[:, 'guess'] = dtc.predict(all_features)
-
-        # Also store the guesses as a synthetic feature
-        sf_hash = '-'.join(sorted(input_df.columns.values))
-        sf_hash += 'DT-{}-{}'.format(max_features, max_depth)
-        sf_identifier = 'SyntheticFeature-{}'.format(hashlib.sha224(sf_hash.encode('UTF-8')).hexdigest())
-        input_df.loc[:, sf_identifier] = input_df['guess'].values
-
-        return input_df
+        return train_model_and_predict(input_df, DecisionTreeClassifier, max_features=max_features, max_depth=max_depth, random_state=42)
 
     @staticmethod
     def random_forest(input_df, num_trees, max_features):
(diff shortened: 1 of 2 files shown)


tpot/tpot.py
```diff
@@ -724,41 +724,17 @@ def decision_tree(input_df, max_features, max_depth):
...
-        input_df.loc[:, 'guess'] = dtc.predict(all_features)
-
-        # Also store the guesses as a synthetic feature
-        sf_hash = '-'.join(sorted(input_df.columns.values))
-        sf_hash += 'DT-{}-{}'.format(max_features, max_depth)
-        sf_identifier = 'SyntheticFeature-{}'.format(hashlib.sha224(sf_hash.encode('UTF-8')).hexdigest())
-        input_df.loc[:, sf_identifier] = input_df['guess'].values
-
-        return input_df
+        return TPOT._train_model_and_predict(input_df, DecisionTreeClassifier, max_features=max_features, max_depth=max_depth, random_state=42)
```
rhiever: I think the `TPOT.` is extraneous: It should be possible to call `_train_model_and_predict()` without it.

tpot/tpot.py
```diff
@@ -724,41 +724,17 @@ def decision_tree(input_df, max_features, max_depth):
...
-    def random_forest(input_df, num_trees, max_features):
+    def random_forest(input_df, n_estimators, max_features):
         """Fits a random forest classifier
 
         Parameters
         ----------
         input_df: pandas.DataFrame {n_samples, n_features+['class', 'group', 'guess']}
             Input DataFrame for fitting the random forest
-        num_trees: int
+        n_estimators: int
```
rhiever: Fine with me. :+1:

tpot/tpot.py
```diff
@@ -979,31 +868,63 @@ def gradient_boosting(input_df, learning_rate, n_estimators, max_depth):
...
-        
+
         input_df = input_df.copy()
 
         training_features = input_df.loc[input_df['group'] == 'training'].drop(['class', 'group', 'guess'], axis=1).values
         training_classes = input_df.loc[input_df['group'] == 'training', 'class'].values
+       
+        # Try to seed the random_state parameter if the model accepts it.
+        try:
+            clf = model(random_state=42,**kwargs)
```
rhiever: I'm pondering if seeding the random state is even necessary. Maybe it's even harmful in the long run?

This isn't something we need to address here, but I've been considering exploring that more.

rhiever:
Great work as always! Thank you for adding the test.

Addressing your questions:

>1) Maybe it's just me, but It feels a little odd to call a static function from another static function. Does it make sense to move this method (and perhaps some others) to a utility function module? Either way, is it faster to call this abstract method as a static method inside the class, a class method inside the class, or something else?

AFAIK, there is no performance difference between the two. The difference between a static and regular class function is whether the function relies on the current state of the object, i.e., it affects the scope. Static functions always have their own scope, whereas class functions share the scope of the object they're called on.

I think we have to ask ourselves: Do we want users using these functions outside of TPOT? e.g.,

```python
from tpot import random_forest

result = random_forest(...)
...
```

I originally envisioned the `export()` function doing something like that, but I ended up exporting the pipelines directly to sklearn code (even if it looks ugly). Perhaps we should change all of the functions to regular class methods since we don't really want users doing that. I don't think any of the functions in TPOT were really made to be used outside of TPOT.

>2) Is there a more appropriate name for the method?

Name looks fine to me!

>3) Is there a better place to do the input validation (i.e. if there are only three columns in the input_df) than within this nested function call?

That's a tough one, since we're never really sure where the pipeline operator functions will be getting their input from and where they're sending it to. I think putting in that quick check at the beginning of the function is the best we can do.

rhiever:
IMO we should keep `_train_model_and_predict()` within the TPOT class. It's simply an abstraction of the  TPOT classifier functions.

bartleyn:
I settled on moving `_train_model_and_predict()` out of the class to see if we could remove the `TPOT.` that prefaced the function call in the static models; I agree that ideally the function should remain in the TPOT class. However, without passing an instantiated TPOT object into the static model call, I'm not sure how to avoid the static `TPOT._train_model_and_predict()`.

rhiever:
I see. We should definitely remove all of the `@staticfunction` decorators then, and add `self` as the first parameter to these functions. That will allow you to call the internal functions as `self.function_name()`.

bartleyn:
I'll try to squash some of my commits for brevity's sake.

rhiever:
This looks ready to merge now. Anything else you planning on adding to this PR?

bartleyn:
I think I'm good for now. If you think it's ready, then great!

Issue #44, opened 2015-12-05
Break export() down into 3 separate functions


rhiever:
`export()` is currently too large. Break it down into 3 functions based on the primary steps of the function:

- Replace all of the mathematical operators with their results

- Unroll the nested function calls into serial code

- Replace the function calls with their corresponding Python code

This change will make it easier to unit test the `export()` function as well.

UniqueFool:
also, the website/example may need to be updated to ensure that the examples there actually work with what's currently available via pip

pronojitsaha:
@rhiever While I am working on this, can you provide some inputs on the 'type' and 'contents' of the following just to be sure
-exported_pipeline, 
-pipeline_list, and 
-operator_text

rhiever:
@pronojitsaha, it may be best to hold off on this issue until we see what affect @dmarx's refactor (see https://github.com/rhiever/tpot/issues/43#issuecomment-165660908) has on the export function. I suspect it will shrink the export function significantly.

pronojitsaha:
Ok. I see that @dmarx's implementation still has the first two functions in export i.e. 

- Replace all of the mathematical operators with their results

- Unroll the nested function calls into serial code

Should I break that into separate functions, it will streamline export() further and then incorporate @dmarx's refactor?

dmarx:
@pronojitsaha: Sounds good to me

pronojitsaha:
@dmarx Ok..great! Will get it done tomorrow.

rhiever:
:+1:

pronojitsaha:
@rhiever While we await the integration of #63, it will be good to break the final part i.e. 'Replace the function calls with their corresponding Python code' into a separate function or a code that the main tpot file imports from for improved code readability and comprehensibility significantly (similar way that I did the first two). I already have a basic structure, if you want can create a PR for it and you can look at it?

rhiever:
Sounds good to me.

On Wednesday, December 23, 2015, PRONOjit Saha <notifications@github.com>
wrote:

> @rhiever <https://github.com/rhiever> While we await the integration of
> #63 <https://github.com/rhiever/tpot/pull/63>, it will be good to break
> the final part i.e. 'Replace the function calls with their corresponding
> Python code' into a separate function or a code that the main tpot file
> imports from for improved code readability and comprehensibility
> significantly (similar way that I did the first two). I already have a
> basic structure, if you want can create a PR for it and you can look at it?
>
> —
[...]

PR #62, opened 2015-12-19
Break export() into sub-functions

pronojitsaha:
Build separate functions for the following steps of `export()`:

-Replace all of the mathematical operators with their results

-Unroll the nested function calls into serial code

Closes issue #44, while awaiting implementation of issue #43


rhiever:
Thank you @pronojitsaha!

PR #66, opened 2015-12-24
Break export() into sub-parts

pronojitsaha:
Implements the final part of #44 i.e. 'Replace the function calls with their corresponding Python code'

### DIFF
--- a/tpot/export_utils.py
+++ b/tpot/export_utils.py
@@ -1,9 +1,91 @@
-def _replace_function_calls(pipeline_list):
+# Utility functions that convert the current optimized pipeline into its corresponding python code
+# For usage see export() function in tpot.py
 
-    # Replace the function calls with their corresponding Python code
-    # For usage see export() function in tpot.py
+import deap
+
+
+def replace_mathematical_operators(exported_pipeline):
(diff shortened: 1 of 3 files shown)


rhiever:
It looks like the build for Python 3.4 is failing. Maybe something to do with the new module?

Also, perhaps we should rename `tpot_export.py` to `export_utils.py` and move all of the export-related functions there (except for `export()`, of course).

pronojitsaha:
Yes I saw that too, the error is in importing the new module. `ERROR: Failure: ImportError (No module named 'tpot_export')`. In my local machine it runs perfectly. Can you run the build once again and check?

Else will check this over the weekend. Renaming to `export_utils.py` is a good idea. :+1: Will implement that too. 

And lastly but most importantly, Merry Christmas :christmas_tree: to you and also to @rasbt and everyone else!

rhiever:
I think you need to change:

```python
from tpot_export import ...
```

to

```python
from .tpot_export import ...
```

to clarify that it's a local import.

Merry Christmas all! :santa:

pronojitsaha:
@rhiever Thanks for the tip! I think this looks good now. Let me know.

Issue #58, opened 2015-12-15
Replace PCA operator with RandomizedPCA


rhiever:
RandomizedPCA is much faster (especially on large data sets) and achieves approximately the same result.

PR #72, opened 2016-01-08
Changing PCA to RandomizedPCA

kadarakos:
Solving issue #58, since the default value (3) for the number of iterations for the power method  performs well in most cases I made the decision to threshold it between [1, 10] keeping it in the same order of magnitude.


rhiever:
Looks great - thank you!

PR #73, opened 2016-01-19
Documenting Pipeline Operators / Public Functions

bartleyn:
This is intended to be an enhancement to the documentation of the pipeline operators & public functions of the main TPOT class. Naturally this is just a first stab at formatting the pages, but hopefully we will not have to change much for the big refactor. At the very least there is some content for each of the operators & functions. I've also changed the theme and mkdocs config to help incorporate navigation to the new reference.

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -11,8 +11,6 @@ theme: readthedocs
 markdown_extensions:
   - tables
   - fenced_code
-  - toc:
-      permalink: []
 
 copyright: Copyright &copy; 2015 <a href="http://randalolson.com">Randal Olson</a>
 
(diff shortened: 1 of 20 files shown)


rhiever:
>This branch has conflicts that must be resolved

Know what's causing that conflict?

bartleyn:
My mkdocs.yml was the source of the conflict. Seems resolved.

rhiever:
Great start! I think the "example exported code" would also need to include the import statements for the corresponding operator, right?

bartleyn:
You're totally right. I reused the first part for all of the examples, and neglected to update import statements. I'll throw those in when I get a chance.

rhiever:
Also please note that in #72 PCA was changed to RandomizedPCA. The operator is still named `_pca`, but it calls RandomizedPCA instead.

bartleyn:
Changes should be integrated.

rhiever:
Merged - thanks again!

PR #88, opened 2016-02-19
Replace scikit GB with XGB in docs

tcfuji:
## What does this PR do?
Replaces the scikit GB classifier with the XGB classifier in the docs.

mkdocs.yml, ref_index.md, and XGBClassifier.md. Also made a minor change in tpot.py.

Don't think so.

This fixes what was discussed in the previous XGBoost PR. #83 

#83 

## Questions:

They are being updated in this PR.
No


rhiever:
Thanks @tcfuji, you rock!

Issue #99, opened 2016-03-01
Use sklearn train_test_split in place of StratifiedShuffleSplit


rhiever:
The latest version of sklearn's [train_test_split](http://scikit-learn.org/stable/modules/generated/sklearn.cross_validation.train_test_split.html) automatically stratifies the data by class(es) and shuffles the data prior to the split. Thus, we don't need to use the long-form version with [StratifiedShuffleSplit](http://scikit-learn.org/stable/modules/generated/sklearn.cross_validation.StratifiedShuffleSplit.html).

Let's replace all of the StratifiedShuffleSplit calls with train_test_split.

pronojitsaha:
Ok..will look into this.

rhiever:
Ping. How's this coming along? Should I mark it as "being worked on"?

pronojitsaha:
Yes, I did give it a try couple of days back, but getting some spurious results. Checking on it. yes you can mark it.

pronojitsaha:
Have sent across the PR. Thanks.

PR #112, opened 2016-03-12
Change StratifiedShuffleSplit to ttrain_test_split

pronojitsaha:
Addresses #99


rhiever:
Hmmm... it's saying that no changes were made. Did you update from `master` and overwrite your changes?

pronojitsaha:
In hurry, forgot to commit! Should be good now.

rhiever:
Thanks! Did you verify that it returns the same splits? If not, we'll have to verify that before merging.

rhiever:
Just checked and, by default, `train_test_split` doesn't stratify the data by class. You have to pass the `stratify` option and a list of the class labels, e.g.,

```python
X_train, X_test, y_train, y_test = train_test_split(input_data.drop('class', axis=1).values, 
                                                    input_data['class'].values,
                                                    train_size=0.75, test_size=0.25,
                                                    random_state=RANDOM_STATE,
                                                    stratify=input_data['class'].values)
```

pronojitsaha:
Ok..sure.

pronojitsaha:
I have checked the splits, and the split ratio is same for both `train_test_split` and `StratifiedShuffleSplit`, but the split indices are somewhat different which is expected.

rhiever:
If you set the `random_state` to the same thing in your tests, they should come out the same. That's what I verified on my end yesterday.

tpot/tpot.py
```diff
@@ -214,10 +214,10 @@ def fit(self, features, classes):
             np.random.shuffle(data_columns)
             training_testing_data = training_testing_data[data_columns]
 
-            training_indices, testing_indices = next(iter(StratifiedShuffleSplit(training_testing_data['class'].values,
-                                                                                 n_iter=1,
+            training_indices, testing_indices = train_test_split(training_testing_data.index,
```
rhiever: This doesn't look right. The call needs to look something like:

```python
(training_features, testing_features,
training_labels, testing_labels) = train_test_split(input_data.drop('class', axis=1).values, 
                                                    input_data['class'].values,
                                                    train_size=0.75, test_size=0.25,
                                                    random_state=RANDOM_STATE,
                                                    stratify=input_data['class'].values)
```

Have you tested this?

pronojitsaha: Yes, I have tested it on IRIS and MNIST dataset. Works the same. We can also do it the way you have pointed out, but using `training_testing_data.index` to get `training_indices, testing_indices` is in line with rest of our code format.

pronojitsaha:
Ok, got that. Thanks.

rhiever:
Understood. Alright, looks good to merge. Thanks again! :-)

Issue #114, opened 2016-03-17
Can u add an example where we can use the data in numpy for the generated "tpot_mnist_pipeline.py"


anjith2006:
In the example MNIST, the generated python code is not directly usable for evaluation with the original 
data. 
Can you post an example were I can directly use the scikit type , or simply numpy arrays for data and labels.

rhiever:
Do you mean [this example](https://github.com/rhiever/tpot#example)? Here's an example of that code transforming the data from `load_digits()` into pandas DataFrame format.

```python
import numpy as np
import pandas as pd

from sklearn.cross_validation import StratifiedShuffleSplit
from sklearn.linear_model import LogisticRegression
from sklearn.datasets import load_digits

digits = load_digits()

tpot_data = pd.DataFrame(digits.data)
tpot_data['class'] = digits.target

training_indeces, testing_indeces = next(iter(StratifiedShuffleSplit(tpot_data['class'].values, n_iter=1, train_size=0.75, test_size=0.25)))

result1 = tpot_data.copy()

# Perform classification with a logistic regression classifier
lrc1 = LogisticRegression(C=2.8214285714285716)
lrc1.fit(result1.loc[training_indeces].drop('class', axis=1).values, result1.loc[training_indeces, 'class'].values)
result1['lrc1-classification'] = lrc1.predict(result1.drop('class', axis=1).values)
```

pronojitsaha:
Hey @rhiever , we need to change the `StratifiedShuffleSplit` to `train_test_split` in the above tpot example as well.

rhiever:
Sure, submit the PR? :-)

On Thursday, March 17, 2016, PRONOjit Saha <notifications@github.com> wrote:

> Hey @rhiever <https://github.com/rhiever> , we need to change the
> StratifiedShuffleSplit to train_test_split in the above tpot example as
> well.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/rhiever/tpot/issues/114#issuecomment-197866047>
>


-- 
Randal S. Olson, Ph.D.
Postdoctoral Researcher, Institute for Biomedical Informatics
University of Pennsylvania

E-mail: rso@randalolson.com | Twitter: @randal_olson
<https://twitter.com/randal_olson>
http://www.randalolson.com

pronojitsaha:
OK...will look into it and send across the PR.

PR #116, opened 2016-03-19
Update Example section of Readme.md

pronojitsaha:
@rhiever Readme.md has the tpot_mnist_pipeline.py (in the Example section) which is of a very early version. We have updated the tpot_mnist_pipeline.py in the tutorials section..but not done so in the Example of Readme.md. So I have updated the same including changing `StratifiedShuffleSplit` to `train_test_split` as discussed in #114


rhiever:
I think all the extra models makes the example code look too complex. How about we just have the example code be a generated Random Forest call with 100+ estimators? That's usually what comes out of TPOT for the MNIST data set now anyway.

pronojitsaha:
Yes, I was also thinking about the complexity to understand this for a first time user. I will look into it.

rhiever:
Good to go! Cheers.

Issue #130, opened 2016-04-18
Add more preprocessing operators


rhiever:
I have listed more preprocessing operators that we can add from sklearn.

# FastICA

http://scikit-learn.org/stable/modules/generated/sklearn.decomposition.FastICA.html

Evolvable parameters: n_components, tol

Fixed parameters: random_state = 42

# FeatureAgglomeration

http://scikit-learn.org/stable/modules/generated/sklearn.cluster.FeatureAgglomeration.html

Evolvable parameters: n_clusters, affinity, linkage

Fixed parameters: none

# Kernel PCA

http://scikit-learn.org/stable/modules/generated/sklearn.decomposition.KernelPCA.html

Evolvable parameters: n_components, kernel, gamma, coef0, tol

Fixed parameters: none

# Nystroem

http://scikit-learn.org/stable/modules/generated/sklearn.kernel_approximation.Nystroem.html

Evolvable parameters: kernel, n_components, gamma, coef0

Fixed parameters: random_state = 42

# RandomTreesEmbedding

http://scikit-learn.org/stable/modules/generated/sklearn.ensemble.RandomTreesEmbedding.html

Evolvable parameters: none

Fixed parameters: n_estimators = 500, random_state = 42, n_jobs = -1

PR #132, opened 2016-04-25
New preprocessors

teaearlgraycold:
## What does this PR do?

Adds a few new feature preprocessors to TPOT:

* FastICA
* FeatureAgglomeration
* Nystroem

(RandomTreesEmbedding was found to increase the feature count too much and was not added)

Line 1143 of tpot.py, where the new operators are implemented.

The preprocessors should be easily tested by commenting out most other pipeline operators and running TPOT on a dataset, forcing TPOT to use these new operators. I would also recommend exporting the pipeline to sklearn code.

About the Nystroem preprocessor, I mistakenly started working on it when @rhiever had said not to, misreading one of his messages. But I believe his reasoning for not wanting it was that it required too many parameters for GP to be able to optimize.

But the sklearn docs seem to indicate that one of the (optional) parameters isn't even used by the majority of the kernel types, so I left it out of the operator code. So now Nystroem only requires 3 parameters.

#130

## Questions:

I have already updated them

No


rhiever:
Needs unit tests, then we can merge.

teaearlgraycold:
Also added operator that adds features for the count of zero and non-zero elements as per #133

rhiever:
Hrm. This branch has conflicts now because of the export_utils refactor. Should be a small fix, yes?

teaearlgraycold:
Sorry, had forgotten to push that commit until now. Should be ready to merge in.

docs/sources/documentation/pipeline_operators/cluster/FeatureAgglomeration.md
```diff
@@ -0,0 +1,29 @@
...
+        Input integer is used to select one of the above strings.
+    linkage: int
+        Can be one of the following values:
+            "ward", "complete", "average"
+        Input integer is used to select one of the above strings.
+
+Returns
+-------
+    modified_df: pandas.DataFrame {n_samples, n_components + ['guess', 'group', 'class']}
+        Returns a DataFrame containing the transformed features
```
rhiever: Needs an example export case.

tpot/tpot.py
```diff
@@ -1135,6 +1141,210 @@ def _rbf(self, input_df, gamma, n_components):
...
+
+        """
+        feature_cols_only = input_df.drop(self.non_feature_columns, axis=1)
+
+        num_features = len(feature_cols_only.columns.values)
+
+        if num_features == 0:
+            return input_df.copy()
+
+        non_zero_col = np.array([np.count_nonzero(row) for i, row in feature_cols_only.iterrows()]).astype(np.float64)
```
rhiever: This should be doable without a `for` loop, which will speed things up on larger data sets. Something like:

```python
    feature_cols_only.apply(lamda row: np.count_nonzero(row), axis=1)
```

tpot/tpot.py
```diff
@@ -1135,6 +1141,210 @@ def _rbf(self, input_df, gamma, n_components):
...
+        kernel_name = kernel_types[kernel % len(kernel_types)]
+
+        nys = Nystroem(kernel=kernel_name, gamma=gamma, n_components=n_components)
+        nys.fit(training_features.values.astype(np.float64))
+        transformed_features = nys.transform(input_df.drop(self.non_feature_columns, axis=1).values.astype(np.float64))
+
+        modified_df = pd.DataFrame(data=transformed_features)
+        modified_df['class'] = input_df['class'].values
+        modified_df['group'] = input_df['group'].values
+        modified_df['guess'] = input_df['guess'].values
```
rhiever: In all places where we do this: I wonder if it should be replaced by a `for` loop that does this:

```python
for non_feature_column in self.non_feature_columns:
    modified_df[non_feature_column] = input_df[non_feature_column].values
```

in case we add or remove non-feature columns in the future.

teaearlgraycold:
Well both of those would end up being O(n) operations. Would using the `apply()` method be faster because of use of C speedups?

docs/sources/documentation/pipeline_operators/decomposition/FastICA.md
```diff
@@ -0,0 +1,60 @@
...
+```Python
+import numpy as np
+import pandas as pd
+from sklearn.cross_validation import StratifiedShuffleSplit
+from sklearn.tree import DecisionTreeClassifier
+from sklearn.decomposition import FastICA
+
+# NOTE: Make sure that the class is labeled 'class' in the data file
+tpot_data = pd.read_csv('PATH/TO/DATA/FILE', sep='COLUMN_SEPARATOR')
+training_indices, testing_indices = next(iter(StratifiedShuffleSplit(tpot_data['class'].values, n_iter=1, train_size=0.75, test_size=0.25)))
```
rhiever: All of the docs need to be updated with the new way we implement the cross-validation split. We use `train_test_split` now, e.g.,

```python
training_indices, testing_indices = train_test_split(tpot_data.index, stratify=tpot_data['class'].values, train_size=0.75, test_size=0.25)
```

tpot/tpot.py
```diff
@@ -1135,6 +1141,210 @@ def _rbf(self, input_df, gamma, n_components):
...
+        non_zero_col = np.array([np.count_nonzero(row) for i, row in feature_cols_only.iterrows()]).astype(np.float64)
+        zero_col     = np.array([(num_features - x) for x in non_zero_col]).astype(np.float64)
+
+        modified_df = feature_cols_only.copy()
+        modified_df['non_zero'] = pd.Series(non_zero_col, index=modified_df.index)
+        modified_df['zero_col'] = pd.Series(zero_col, index=modified_df.index)
+
+        modified_df['class'] = input_df['class'].values
+        modified_df['group'] = input_df['group'].values
+        modified_df['guess'] = input_df['guess'].values
```
rhiever: All of this shouldn't be necessary here. You should only need to assign the new columns, e.g.,

```python
modified_df['non_zero'] = np.array([np.count_nonzero(row) for i, row in feature_cols_only.iterrows()]).astype(np.float64)
modified_df['zero_col'] = np.array([(num_features - x) for x in non_zero_col]).astype(np.float64)
```

(or however we end up computing zero vs. non-zero entries, per my other comment)

tpot/tpot.py
teaearlgraycold: I'd like to be able to factor that into a function somewhere so we have less copy/pasted code.

```python
def example_preprocessor(self, input_df):
    . . .
    
    return copy_nfc(modified_df, input_df).copy()

def copy_nfc(self, m_df, in_df):
    for non_feature_column in self.non_feature_columns:
        m_df[non_feature_column] = in_df[non_feature_column].values

    # Maybe throw this in there, too?
    new_col_names = {}
    for column in m_df.columns.values:
        if type(column) != str:
            new_col_names[column] = str(column).zfill(10)
    m_df.rename(columns=new_col_names, inplace=True)

    return m_df
```

tpot/tpot.py
```diff
@@ -1135,6 +1141,210 @@ def _rbf(self, input_df, gamma, n_components):
...
+
+        new_col_names = {}
+        for column in modified_df.columns.values:
+            if type(column) != str:
+                new_col_names[column] = str(column).zfill(10)
+        modified_df.rename(columns=new_col_names, inplace=True)
```
rhiever: This code shouldn't be necessary either since we're not reshaping the feature space, only adding a couple extra features.

tpot/tpot.py
```diff
@@ -1135,6 +1141,210 @@ def _rbf(self, input_df, gamma, n_components):
...
+
+        if len(training_features.columns.values) == 0:
+            return input_df.copy()
+
+        if n_components < 1:
+            n_components = 1
+        else:
+            n_components = min(n_components, len(training_features.columns.values))
+
+        kernel_types = list(PAIRWISE_KERNEL_FUNCTIONS.keys())
```
rhiever: I think we should hard-code this list as we do in `_feat_agg()`. I'm concerned that the order may change in future versions of sklearn, and that could mess up our encoding here. Instead, let's take the current values/ordering from `list(PAIRWISE_KERNEL_FUNCTIONS.keys())` and hard-code it.

rhiever:
Reviewed in detail & commented on this PR. Let me know your thoughts, @teaearlgraycold.

rhiever:
>Well both of those would end up being O(n) operations. Would using the `apply()` method be faster because of C speedups?

Yes, I believe so. I've typically found speedups from replacing `for` loops with `apply()` calls on pandas DataFrames.

tpot/tpot.py
teaearlgraycold: Well the class column is a float (after data cleaning/preprocessing), right? I don't want to count it as a column with a value of zero/non-zero.

tpot/tpot.py
teaearlgraycold: You're right

tpot/tpot.py
rhiever: Now I wonder: Do we have enough preprocessor functions with common code that we should implement a generic preprocessor function equivalent to `_train_model_and_predict`?

tpot/tpot.py
rhiever: `class`, `group`, and `guess` should be dropped out in the `feature_cols_only` DataFrame, right?

tpot/tpot.py
teaearlgraycold: Sounds good to me

tpot/tpot.py
teaearlgraycold: They're dropped out, so I add them to the returned DataFrame with lines 1336 - 1338

rhiever: Right, but that shouldn't be necessary. It's less work to do, e.g.,

```python
modified_df = input_df.copy()
modified_df['non_zero'] = np.array([np.count_nonzero(row) for i, row in feature_cols_only.iterrows()]).astype(np.float64)
modified_df['zero_col'] = np.array([(num_features - x) for x in non_zero_col]).astype(np.float64)
```

teaearlgraycold: Oh, I'd misunderstood. That should work just fine and be simpler.

docs/sources/documentation/pipeline_operators/decomposition/FastICA.md
```diff
@@ -0,0 +1,61 @@
...
+        Returns a DataFrame containing the transformed features
+
+
+Example Exported Code
+---------------------
+
+```Python
+import numpy as np
+import pandas as pd
+from sklearn.cross_validation import StratifiedShuffleSplit
```
rhiever: Don't forget the imports for the `train_test_split` docs change. :-)

teaearlgraycold:
There's also a reference to StratifiedShuffleSplit in `tutorials/Titanic_Kaggle.ipynb`. Not sure if you want that changed.

rhiever:
>There's also a reference to StratifiedShuffleSplit in tutorials/Titanic_Kaggle.ipynb. Not sure if you want that changed.

Sure, let's remove all of them while we're at it. Thank you for going through and cleaning up the docs like this.

Issue #140, opened 2016-05-10
Add progress bar


stared:
Since it takes quite a lot of time to get the results: is there any way to use it with a progress bar, e.g. [tqdm](https://github.com/tqdm/tqdm) (For example, per generation; `.next()` method would suffice)? 

Or even better, a built-in progress bar as in [PyMC3](https://pymc-devs.github.io/pymc3/) (for an example: https://github.com/markdregan/Bayesian-Modelling-in-Python/blob/master/Section%203.%20Hierarchical%20modelling.ipynb).

rhiever:
Interesting idea. We currently print the progress every generation of the algorithm, but perhaps we could also print the progress toward evaluating each generation's population along the way too. I certainly appreciate having a tqdm-like progress bar for long-running for loops.

stared:
Now I see that it prints progress per step (with `verbosity=2` or all options?). 
Yet, since my computer is relatively old, I though it doesn't (only after a few minutes on the `hill_valley_noisy_data` example I got he first step; I didn't know if it was running at all).

Or is the initialization particularly  costly? (The next steps were much shorter)

rhiever:
It only prints the progress per generation with `verbosity=2`. I was aiming for a "quiet" classifier by default.

>Or is the initialization particularly costly? (The next steps were much shorter)

The initialization shouldn't be much more expensive than every other generation. Perhaps what happened was your initial population had a complex pipeline in it that takes a long time to evaluate, and the selection procedure for the next TPOT generation (that aims to maximize accuracy but minimize pipeline complexity) threw that slow pipeline out because it was too complex. That's my best guess.

teaearlgraycold:
I'm a big fan of this idea (and tqdm in general).

rhiever:
We'll have to figure out how to get tqdm wrapped into the evolutionary algorithm. As we discussed yesterday, @teaearlgraycold, the GP algorithm is all in the `eaSimple` call. Here's the eaSimple code: https://github.com/DEAP/deap/blob/master/deap/algorithms.py#L162

teaearlgraycold:
Unless we're going to monkey-patch over DEAP I think we should consider the number of steps in the process of fitting a classifier pipeline to be equal to the population count multiplied by the generation count.

Then we can use the manual approach to tqdm's stepping like so:

```python
def fit(self, features, classes):
    total = self.population_size * self.generations

    with tqdm(total=total) as self.pbar:
        # Continue with remainder of function

[...]

PR #144, opened 2016-05-11
Add support for progress bar during GP if verbosity is set to 2

teaearlgraycold:
## What does this PR do?

Adds a progress bar (from the tqdm module) that displays during calls to fit() if verbosity is set to 2

The reviewer should check the fit() and _gp_new_generation functions

Running TPOT with verbosity set to 2

#140

As you can see below there's a bit of a conflict between the GP min/max/avg stats and the tqdm progress bar. Typically tqdm will smoothly update the progress bar as it progresses, but the stat output forces tqdm to redraw the bar. So it may be desirable to either remove the stats or change the conditions under which the stats and tqdm show up.

Console-shot:

```
GP Progress:  22%|██████▌                      | 9/40 [00:01<00:06,  4.98pipeline/s]
gen	nevals	Minimum score	Average score	Maximum score
0  	10    	0.5          	0.662568     	0.972352                                                         
GP Progress:  50%|██████████████              | 20/40 [00:06<00:09,  2.14pipeline/s]
1  	10    	0.5          	0.837614     	0.984079                                                                
GP Progress:  70%|███████████████████▌        | 28/40 [00:11<00:07,  1.63pipeline/s]
2  	8     	0.97224      	0.979522     	0.984079                                                            
GP Progress:  95%|██████████████████████████▌ | 38/40 [00:19<00:01,  1.15pipeline/s]
3  	8     	0.881496     	0.968862     	0.984079  
```

**Edit:**

Above issue is no longer present in the branch

## Questions:

There does not seem to be any information in the docs about what the verbosity setting does at the moment, so I suppose not unless it is desirable to add that now.

Yes, `tqdm`


teaearlgraycold:
Sadly I don't think it's easy to test this much more without running GP during the tests (ew), or displaying the progress bar in stdout during tests (also bad IMO).

teaearlgraycold:
So I removed the stats that were set up with DEAP and did a bit of refactoring on how I defined the _gp_new_generation decorator based on what StackOverflow seemed to indicate was the proper way to define this type of decorator (having it at the module level).

Since it seems decorators need to be defined *before* they are called I opted to add a decorators file to import _gp_new_generation from rather than keep it permanently above the TPOT class definition in tpot.py.

tpot/tpot.py
```diff
@@ -261,16 +267,29 @@ def pareto_eq(ind1, ind2):
...
             self.hof = tools.ParetoFront(similar=pareto_eq)
 
-            stats = tools.Statistics(lambda ind: ind.fitness.values[1])
-            stats.register('Minimum score', np.min)
-            stats.register('Average score', np.mean)
-            stats.register('Maximum score', np.max)
+            # stats = tools.Statistics(lambda ind: ind.fitness.values[1])
+            # stats.register('Minimum score', np.min)
+            # stats.register('Average score', np.mean)
+            # stats.register('Maximum score', np.max)
```
rhiever: Why comment these out? I think we'd just remove them. They're in the git history if we ever want these lines back.

tpot/tpot.py
```diff
@@ -261,16 +267,29 @@ def pareto_eq(ind1, ind2):
...
 
-            pop, _ = algorithms.eaSimple(population=pop, toolbox=self._toolbox, cxpb=self.crossover_rate,
+            # Start the progress bar
+            num_evaluations = self.population_size * (self.generations + 1)
+            self.pbar = tqdm(total=num_evaluations, unit='pipeline',
+                             disable=(not verbose), desc='GP Progress')
+
+            with warnings.catch_warnings():
+                # Ignore warnings
+                warnings.simplefilter('ignore', category=UserWarning)
```
rhiever: What warnings is this meant to ignore? At this point, I don't want to have a catch-all that throws out all warnings because those warnings may be useful to us and the user.

tpot/tpot.py
```diff
@@ -261,16 +267,29 @@ def pareto_eq(ind1, ind2):
...
+
+                pop, _ = algorithms.eaSimple(population=pop, toolbox=self._toolbox, cxpb=self.crossover_rate,
                                          mutpb=self.mutation_rate, ngen=self.generations,
-                                         stats=stats, halloffame=self.hof, verbose=verbose)
+                                         halloffame=self.hof, verbose=False)
+
+            # Close the progress bar
+            self.pbar.close()
```
rhiever: What happens to the progress bar if the run is interrupted early?

tpot/tpot.py
```diff
@@ -261,16 +267,29 @@ def pareto_eq(ind1, ind2):
...
+            self.gp_generation = 0
```
rhiever: Why does `self.gp_generation` need to be set to 0 to close the progress bar?

tpot_mnist_pipeline.py
```diff
@@ -0,0 +1,17 @@
+import numpy as np
```
rhiever: Was this file intended to be included?

tpot/tpot.py
teaearlgraycold: Some operators (like FastICA) can be noisy during a call to fit().

SKLearn would say that `tol` was too low for convergence, but I assumed the GP would just work that out and there wasn't any need to actually display those types of errors, especially considering how the user has no direct input when it comes to those evolve-able parameters and they mess with the progress bar.

Maybe tol just need to have a minimum allowed value for FastICA. I believe I've seen other operators have similar issues, though.

tpot/tpot.py
teaearlgraycold: Good point, it'd be safer to add the call to `pbar.close()` into a `finally` block after the exception handling.

tpot/tpot.py
rhiever: I see. In that case, I think it's better to discover what specific operators are throwing those warnings and -- if we decide that the warnings should be ignored -- put those "ignore warning" wrapper around the specific calls that throw that warning, as we do in e.g. `select_kbest()`.

In any case, this bit of code should be in a separate issue/PR IMO. Please remove this code from this PR, raise an issue for this concern, and we'll decide on course of action there.

tpot/tpot.py
teaearlgraycold: It doesn't.

I was just returning the state of the TPOT instance to like it was before fit() was called.

That way consecutive calls to fit() don't behave differently.

tpot/tpot.py
rhiever: While we're at it, please move this code down to a `finally` block as well: https://github.com/rhiever/tpot/blob/master/tpot/tpot.py#L275

and remove the duplicate code in `fit`.

tpot/tpot.py
rhiever: Gotcha. Please add a comment explaining that.

tpot/tpot.py
teaearlgraycold: Should I remove the (seemingly unnecessary) `print('')` statement?

https://github.com/rhiever/tpot/blob/master/tpot/tpot.py#L284

rhiever: IIRC the `print('')` is to ensure that there is space between the final outputs and the logs. But if we're removing the logs and replacing them with a progress bar, I think we don't need the extra newline any more.

rhiever:
This PR has merge conflicts now. Pretty sure it has something to do with the argparse help.

teaearlgraycold:
Alright, looks good now

rhiever:
Got an error:

```
GP Progress:   0%|                                                                                | 0/10100 [00:00<?, ?pipeline/s]
n_components is too large: it will be set to 681
GP Progress:   1%|▍                                                                      | 63/10100 [00:14<42:15,  3.96pipeline/s]
n_components is too large: it will be set to 681
GP Progress:   1%|▋                                                                    | 97/10100 [00:37<1:46:22,  1.57pipeline/s]
Best pipeline: _logistic_regression(ARG0, 0.001)
GP Progress:   1%|▋                                                                   | 101/10100 [00:37<1:02:00,  2.69pipeline/s]
  File "/Users/randal_olson/anaconda/lib/python3.5/runpy.py", line 170, in _run_module_as_main
    "__main__", mod_spec)
  File "/Users/randal_olson/anaconda/lib/python3.5/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/Users/randal_olson/Downloads/tpot-progress_bar/tpot/tpot.py", line 1616, in <module>
    main()
  File "/Users/randal_olson/Downloads/tpot-progress_bar/tpot/tpot.py", line 1605, in main
    tpot.fit(training_features, training_classes)
  File "/Users/randal_olson/Downloads/tpot-progress_bar/tpot/tpot.py", line 279, in fit
    halloffame=self.hof, verbose=False)
  File "/Users/randal_olson/anaconda/lib/python3.5/site-packages/deap/algorithms.py", line 162, in eaSimple
    offspring = toolbox.select(population, len(population))
  File "/Users/randal_olson/Downloads/tpot-progress_bar/tpot/decorators.py", line 43, in wrapped_func
    self.pbar.write('Finished generation #{}'.format(self.gp_generation))
AttributeError: 'tqdm' object has no attribute 'write'
```

Also note the warning message, `n_components is too large: it will be set to 681`. That must be a warning related to one of the feature preprocessors.

teaearlgraycold:
What version of tqdm are you running? I've got `4.5.0-a4e00b2`

That's pretty annoying that it's not giving any hint as to which operator is causing that warning. What code are you running?

rhiever:
See my edit. :-)

Anyway, I think the output needs to be cleaned up a bit. Here's what I have for a small TPOT run:

screen shot 2016-05-20 at 9 35 03 am

1) No need to print the message about finishing generations.

2) There should be a newline between the progress bar and the "Best pipeline:" line.

3) Not sure why the progress bar showed up twice?

4) When someone CTRL-C's, the progress bar gets messy. Any way to prevent that?

rhiever:
Actually, is it possible to hide the progress bar entirely when the run finishes (either normally or prematurely)?

teaearlgraycold:
The progress bar should hide itself after a completed run. I think it may double-print when someone force-quits fit(). I'll look into that.

rhiever:
The screenshot I showed was from a run that ended normally, so the progress bar doesn't hide at the end currently.

rhiever:
Now that I think about it: Maybe it would be useful to report the internal CV accuracy of the best pipeline discovered so far at the end of every generation. That way, the user will know when to stop early.

teaearlgraycold:
I don't think that information is accessible from outside the GP. The only way you can access it is by using the stats, right? And that just pumps out information through stdout.

teaearlgraycold:
Oh, and should we somehow specify dependency versions so people use sklearn 0.17, tqdm 4.5.0, etc?

rhiever:
>Oh, and should we somehow specify dependency versions so people use sklearn 0.17, tqdm 4.5.0, etc?

Yes, definitely! We should specify minimum versions required.

rhiever:
>I don't think that information is accessible from outside the GP. The only way you can access it is by using the stats, right? And that just pumps out information through stdout.

The best pipelines should be stored in `self.hof`, and each individual in `self.hof` should have their fitness attached to them.

teaearlgraycold:
Okay. I'll be able to do that very easily then.

teaearlgraycold:
So most of the time the current best score doesn't seem to improve over the first generation.

Maybe that's because I'm doing smaller populations with not many generations (10 - 20 population, 2 - 5 generations). I *am* seeing small improvements some of the time, just not too often.

rhiever:
That's fine. It largely depends on the data set too.

rhiever:
Blargh. This conflicts with the master branch now.

rhiever:
While you sort out the merge conflict (hopefully it's straightforward), please add a newline between the "Best pipeline" and "Generation `x`" outputs, just to make it look a little cleaner.

screen shot 2016-05-25 at 4 23 01 pm

rhiever:
Looks good!

Issue #128, opened 2016-04-18
Add more classifier operators


rhiever:
There are many more classifiers that we can add from sklearn. I have listed them below. We may not want to add some of these classifiers, however, especially the ones that have many evolvable parameters that affect their performance.

Before you proceed to implement a classifier, let's discuss it in this issue.

# AdaBoostClassifier

http://scikit-learn.org/stable/modules/generated/sklearn.ensemble.AdaBoostClassifier.html

Evolvable parameters: learning_rate

Fixed parameters: n_estimators = 500, random_state = 42

# BernoulliNB

http://scikit-learn.org/stable/modules/generated/sklearn.naive_bayes.BernoulliNB.html

Evolvable parameters: alpha, binarize, fit_prior

Fixed parameters: none

# ExtraTreesClassifier

http://scikit-learn.org/stable/modules/generated/sklearn.ensemble.ExtraTreesClassifier.html

Evolvable parameters: criterion, max_features

Fixed parameters: n_estimators = 500, random_state = 42

# GaussianNB

http://scikit-learn.org/stable/modules/generated/sklearn.naive_bayes.GaussianNB.html

Evolvable parameters: none

Fixed parameters: none

# MultinomialNB

http://scikit-learn.org/stable/modules/generated/sklearn.naive_bayes.MultinomialNB.html

Evolvable parameters: alpha, fit_prior

Fixed parameters: none

# LinearSVC

http://scikit-learn.org/stable/modules/generated/sklearn.svm.LinearSVC.html

Evolvable parameters: C, loss, fit_intercept

Fixed parameters: random_state = 42

# PassiveAggressiveClassifier

http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.PassiveAggressiveClassifier.html

Evolvable parameters: C, loss, fit_intercept

Fixed parameters: random_state = 42

# SGDClassifier

http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.SGDClassifier.html

Evolvable parameters: loss, alpha, penalty, fit_intercept, l1_ratio, eta0, power_t, learning_rate

Fixed parameters: random_state = 42

PR #147, opened 2016-05-16
New classifiers

teaearlgraycold:
## What does this PR do?

Adds 7 new classifiers to TPOT

At the _ada_boost() method in the TPOT class, going all the way down to the _p_aggr() method.

There is also relevant export code and docs that should be checked for an LGTM.

Travis should test the classifiers themselves, but a few pipelines could be made and exported to confirm that the code indeed works.

#128

## Questions:

Yes. I have already updated them.

No.

### DIFF
--- a/tpot/tpot.py
+++ b/tpot/tpot.py
@@ -29,7 +29,7 @@ import numpy as np
 import pandas as pd
 
 from sklearn.tree import DecisionTreeClassifier
-from sklearn.ensemble import RandomForestClassifier
+from sklearn.ensemble import RandomForestClassifier, AdaBoostClassifier
 from sklearn.linear_model import LogisticRegression
 from sklearn.svm import SVC
 from sklearn.neighbors import KNeighborsClassifier
@@ -38,6 +38,7 @@ from sklearn.preprocessing import StandardScaler, RobustScaler, MaxAbsScaler, Mi
 from sklearn.preprocessing import PolynomialFeatures, Binarizer
 from sklearn.decomposition import RandomizedPCA
 from sklearn.kernel_approximation import RBFSampler
+from sklearn.naive_bayes import BernoulliNB
 from sklearn.cross_validation import train_test_split
 from xgboost import XGBClassifier
 
@@ -48,11 +49,7 @@ from ._version import __version__
 from .export_utils import *
 
 import deap
-from deap import algorithms
-from deap import base
-from deap import creator
-from deap import tools
-from deap import gp
+from deap import algorithms, base, creator, tools, gp
 
 class TPOT(object):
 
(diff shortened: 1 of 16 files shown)


tpot/tpot.py
```diff
@@ -496,6 +500,181 @@ def _random_forest(self, input_df, max_features):
...
+            Learning rate shrinks the contribution of each classifier by learning_rate.
+
+        Returns
+        -------
+        input_df: pandas.DataFrame {n_samples, n_features+['guess', 'group', 'class', 'SyntheticFeature']}
+            Returns a modified input DataFrame with the guess column updated according to the classifier's predictions.
+            Also adds the classifiers's predictions as a 'SyntheticFeature' column.
+
+        """
+        return self._train_model_and_predict(input_df, AdaBoostClassifier,
```
rhiever: `learning_rate` should be capped at > 0.

Looking at the docs further, I also think that for the AdaBoostClassifier, we should allow `n_estimators` to be evolved as well, with a max of 500 estimators. The AdaBoostClassifier is one unique case where there is a tradeoff between `n_estimators` and `learning_rate`.

tpot/tpot.py
```diff
@@ -496,6 +500,181 @@ def _random_forest(self, input_df, max_features):
...
+            Returns a modified input DataFrame with the guess column updated according to the classifier's predictions.
+            Also adds the classifiers's predictions as a 'SyntheticFeature' column.
+
+        """
+        # Select criterion string from list of valid parameters
+        criterion_values = ['gini', 'entropy']
+        criterion_selection = criterion_values[criterion % len(criterion_values)]
+
+        return self._train_model_and_predict(input_df, ExtraTreesClassifier,
+            criterion=criterion_selection, max_features=max_features,
```
rhiever: `max_features` should be capped between 1 and `len(num_training_features)`.

tpot/tpot.py
```diff
@@ -496,6 +500,181 @@ def _random_forest(self, input_df, max_features):
...
+            Returns a modified input DataFrame with the guess column updated according to the classifier's predictions.
+            Also adds the classifiers's predictions as a 'SyntheticFeature' column.
+
+        """
+        fit_bool = (fit_intercept % 2) == 0
+
+        loss_values = ['hinge', 'squared_hinge']
+        loss_selection = loss_values[loss % len(loss_values)]
+
+        return self._train_model_and_predict(input_df, LinearSVC, C=C,
```
rhiever: `C` should be capped at >0.

tpot/tpot.py
teaearlgraycold: Should I just do

```Python
max(learning_rate, 0.001)
```

Not sure what the exact minimum should be here.

tpot/tpot.py
```diff
@@ -496,6 +500,181 @@ def _random_forest(self, input_df, max_features):
...
+            loss=loss_selection, fit_intercept=fit_bool, random_state=42)
+
+    def _p_aggr(self, input_df, C, loss, fit_intercept):
```
rhiever: Please make this `passive_aggressive()`. `p_aggr` is unclear.

tpot/tpot.py
```diff
@@ -496,6 +500,181 @@ def _random_forest(self, input_df, max_features):
...
+            Returns a modified input DataFrame with the guess column updated according to the classifier's predictions.
+            Also adds the classifiers's predictions as a 'SyntheticFeature' column.
+
+        """
+        fit_bool = (fit_intercept % 2) == 0
+
+        loss_values = ['hinge', 'squared_hinge']
+        loss_selection = loss_values[loss % len(loss_values)]
+
+        return self._train_model_and_predict(input_df, PassiveAggressiveClassifier,
```
rhiever: `C` should be capped at >0.

rhiever:
When writing test cases for classifiers, test with normal parameters as well as extreme parameters: negative values, out-of-bounds values, etc. That will help catch issues where we're allowing invalid parameters to be passed to the various models.

tpot/tpot.py
rhiever: Check `_logistic_regression` for an example: https://github.com/rhiever/tpot/blob/master/tpot/tpot.py#L516

```python
C = max(0.0001, C)
```

0.0001 seems like a fine minimum value for now.

teaearlgraycold:
Now addresses #151

Issue #41, opened 2015-12-04
Expand project unit tests and integration tests


rhiever:
Currently, there are only a few unit tests in `tests.py`. These are basic unit tests and don't cover a large portion of the project. We should expand the unit tests to cover more of the core TPOT functions.

We also need integration tests that test TPOT as a whole. This can be done with a small, fixed data set and a fixed random number generator seed over only a few generations, with a few different parameter settings.

rhiever:
@pronojitsaha, are you interested in taking on this task?

rhiever:
@teaearlgraycold, here's the list of unit tests that @GJena and I brainstormed back when she was working on this issue. Some of them are already implemented.

Unit tests

Run nose tests https://nose.readthedocs.org/en/latest/ Nose/Nose2?

If no features, return copy of data frame..for every feat operator
Df has 3: class, guess, group

tpot.py
Function: `__init__`

Assertion tests present; add more assertions

Cover the new parameters that have been added since the __init__() test was created

Function: fit
fit() can be tested in the integration tests

Function: pareto_eq
Cannot be tested because it is inside fit()

Function: predict
Testing features = training features

Function: score
Use fixed pipeline, RNG, and data sets. Should output the same score each time.

Function: export
Should not be necessary to unit test because all of its functionality will be tested in export_utils.py

Function: decision_tree
Checking if output same as sklearn decision tree

Function: random_forest
Same as sklearn

Function: logistic_regression
Same as sklearn

Function: svc
Same as sklearn

Function: knnc
Same as sklearn

Function: xgradient_boosting
Same as XGBClassifier

Function: train_model_and_predict
Same as sklearn

Function: combine_dfs
input_df1 = input_df2

Function: _rfe
Compare with sklearn

Function: select_percentile
Compare with sklearn

Function: select k_best
Compare with sklearn

Function: select_fwe
Compare with sklearn

Function: variance_threshold
Compare with sklearn

Function: standard_scaler
Compare with sklearn

Function: robust_scaler
Compare with sklearn

Function: polynomial_features
Compare with sklearn

Function: min_max_scaler
Check with features in range
Check with features out of range
If train < test...don’t scale between 0,1..for all feat scaling

Function: max_abs_scaler
Check with features in range
Check with features out of range

Function: binarizer
Check with different thresholds

Function: pca
Compare with sklearn

Function: div remove params, no storing result 
Underflow?
Divide by 0

More verbose comments

Function: evaluate_individual
[...]

PR #157, opened 2016-05-31
Adding new unit tests

teaearlgraycold:
## What does this PR do?

Adds some unit tests to TPOT

tests.py

#41 

---

Should probably just sit on this as I add tests


teaearlgraycold:
Looks like Python 2.7 has different PRNGs.

I should bring up that my testing method of score() isn't amazing anyway, though - I run a call to fit() with a population size of 1 and generation count of zero with TPOT seeded to a specific seed. That way I know exactly what the only pipeline generated should look like.

This however will be affected by the number of operatators and many other factors.

Googling around it seems like I'd want to pickle the deap Individual object. Should we have a bunch of pickled objects ready for use by the test suite?

Also the extra test I added for train_model_and_predict() seems to have issues when it comes to asserting that the testing guesses are identical, so I've commented that out for now.

rhiever:
There's a way to convert a string representation of a pipeline into a
TPOT individual. Let's chat about that tomorrow.

On Tuesday, May 31, 2016, Daniel <notifications@github.com> wrote:

> Looks like Python 2.7 has different PRNGs.
>
> I should bring up that my testing method of score() isn't amazing anyway,
> though - I run a call to fit() with a population size of 1 and generation
> count of zero with TPOT seeded to a specific seed. That way I know exactly
> what the only pipeline generated should look like.
>
> This however will be affected by the number of operatators and many other
> factors.
>
> Googling around it seems like I'd want to pickle the deap Individual
> object. Should we have a bunch of pickled objects ready for use by the test
> suite?
>
> Also the extra test I added for train_model_and_predict() seems to have
> issues when it comes to asserting that the testing guesses are identical,
> so I've commented that out for now.
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/rhiever/tpot/pull/157#issuecomment-222590596>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ABo7t8F3v4zpW55zwha7x-jlCe1BsY_vks5qG7qAgaJpZM4IqKdw>
> .
>


-- 
Randal S. Olson, Ph.D.
Postdoctoral Researcher, Institute for Biomedical Informatics
University of Pennsylvania

E-mail: rso@randalolson.com | Twitter: @randal_olson
<https://twitter.com/randal_olson>
http://www.randalolson.com

teaearlgraycold:
Hey, @rhiever, how would I make a Deap Individual object from a PrimitiveTree.

The following code does seem to create a pipeline that can be evaluated by TPOT.

```Python
tpot_obj._optimized_pipeline = gp.PrimitiveTree.from_string('_random_forest(ARG0, mul(100, 5), 0)', tpot_obj._pset)
```

rhiever:
You have to compile the GP tree representation:

```
self._toolbox.compile(expr=individual)
```

then you should be able to execute the individual as a series of functions.

teaearlgraycold:
Well I'm asking because after running `score()`, and during the subsequent call to `_evaluate_individual()` there's this bit of code:

```Python
# Transform the tree expression in a callable function
func = self._toolbox.compile(expr=individual)

# Count the number of pipeline operators as a measure of pipeline complexity
. . .

result = func(training_testing_data)
```

But with the from_string object used for `_optimized_pipeline`, `result` here has a value of 0, and is not a DataFrame. Normally `_evaluate_individual()` takes an `Individual` object, not a `PrimitiveTree` object.

teaearlgraycold:
Okay, I discovered `deap.creator.Invididual.from_string(string, pset)`. Seems there's some other issue happening in the evaluation method since that did not solve my problem.

teaearlgraycold:
Looks like the string I was pulling from your seed_pipelines module, '_random_forest(ARG0, mul(100, 5), 0)', is not a valid pipeline string.

`_random_forest()` only takes one integer arg.

rhiever:
Yes, that's from an old version of TPOT. :-)

On Tuesday, May 31, 2016, Daniel <notifications@github.com> wrote:

> Looks like the string I was pulling from your seed_pipelines module,
> '_random_forest(ARG0, mul(100, 5), 0)', is not a valid pipeline string.
>
> _random_forest() only takes one integer arg.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/rhiever/tpot/pull/157#issuecomment-222837642>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ABo7t0qjw9yyrOGFvzq58mHsSO7t3qXQks5qHLLFgaJpZM4IqKdw>
> .
>


-- 
Randal S. Olson, Ph.D.
Postdoctoral Researcher, Institute for Biomedical Informatics
University of Pennsylvania

E-mail: rso@randalolson.com | Twitter: @randal_olson
<https://twitter.com/randal_olson>
http://www.randalolson.com

teaearlgraycold:
Somehow we lost the `test_gp_new_generation()` test during a merge or something, so I re-added that.

teaearlgraycold:
Yup, looks like 2.7 uses different PRNG than 3.X

rhiever:
Well how about that.

```
Python 3.5.1 |Anaconda 2.4.0 (x86_64)| (default, Dec  7 2015, 11:24:55)
>>> import random
>>> random.seed(90)
>>> [random.randint(0, 100) for x in range(10)]
[26, 91, 11, 61, 84, 73, 75, 39, 81, 45]
```

```
Python 2.7.11 |Anaconda 2.5.0 (x86_64)| (default, Dec  6 2015, 18:57:58)
>>> import random
>>> random.seed(90)
>>> [random.randint(0, 100) for x in range(10)]
[20, 9, 66, 59, 64, 35, 45, 89, 82, 26]
```

rhiever:
numpy is consistent across versions though:

```
Python 3.5.1 |Anaconda 2.4.0 (x86_64)| (default, Dec  7 2015, 11:24:55)
>>> import numpy as np
>>> np.random.seed(90)
>>> np.random.randint(0, 90, 10)
array([29, 31, 67, 39, 68, 58, 37, 18, 74, 51])
>>> np.__version__
'1.11.0'
```

```
Python 2.7.11 |Anaconda 2.5.0 (x86_64)| (default, Dec  6 2015, 18:57:58)
>>> import numpy as np
>>> np.random.seed(90)
>>> np.random.randint(0, 90, 10)
array([29, 31, 67, 39, 68, 58, 37, 18, 74, 51])
>>> np.__version__
'1.11.0'
```

teaearlgraycold:
http://stackoverflow.com/a/11929775

I'll go ahead and specify the version in __init__

rhiever:
Or just use the numpy RNG?

teaearlgraycold:
The issue I believe is that the libraries we're calling are using random from the standard lib

rhiever:
Okay. Please file it as a bug and fix it ASAP if anywhere in TPOT is using the base `random` library. We'll replace it with the numpy one instead.

I think sklearn and DEAP use the numpy RNG.

tpot/tpot.py
```diff
@@ -1657,13 +1657,16 @@ def _combined_selection_operator(self, individuals, k):
         """
         return tools.selNSGA2(individuals, int(k / 5.)) * 5
 
-    def _random_mutation_operator(self, individual):
+    def _random_mutation_operator(self, individual, mutation_type=None):
```
rhiever: We shouldn't need to change the mutator function to support unit tests. :-)

teaearlgraycold:
Well, that doesn't work:

```
>>> import random
>>> random.seed(90, version=2)
>>> [random.randint(0, 100) for x in range(10)]
[26, 91, 11, 61, 84, 73, 75, 39, 81, 45]
>>> random.seed(90, version=1)
>>> [random.randint(0, 100) for x in range(10)]
[26, 91, 11, 61, 84, 73, 75, 39, 81, 45]
```

tpot/tpot.py
teaearlgraycold: Yeah, but the alternative felt a bit janky - in that I'd need to find 3 separate seeds that would lead me to each branch of the method

rhiever: I understand. I'd rather lose coverage than modify the function to accommodate a unit test.

teaearlgraycold:
Looks like std. lib random is only called once in TPOT, so that should be an easy fix.

teaearlgraycold:
Looks like it's still failing - it's making a call to deap in the tested function, so that's not really a surprise.

rhiever:
Yup, as expected. That's too bad - we just won't test that function.

teaearlgraycold:
If your deap Issue gets recognized and a contributor says it's not a bad idea I'll go ahead and PR the fix - there's only a few hundred references to random in deap, all of which should be easily enough replaced with calls to np.random.

They already use np.random, too - so it wouldn't be adding anything new.

rhiever:
Let's keep an eye on https://github.com/DEAP/deap/issues/138 then

Issue #162, opened 2016-06-03
Remove mathematical operators from the TPOT primitives


rhiever:
Currently, TPOT has mathematical primitives (add, subtract, multiply, divide) that act on integers and floats to decide the parameters of various models and feature preprocessors. I suspect that the mathematical primitives have been causing several issues in TPOT:

1) **They cause bloat.** Since the mathematical operators don't count against the pipeline size (which is part of the Pareto fitness), GP tends to add tons of mathematical operators to the pipelines. This in turn causes most of the mutation and crossover events to affect the mathematical operators rather than exploring different pipeline operators in the pipeline. (The latter has yet to be experimentally verified, but is probable.)

2) **They produce extreme values.** Many of the model parameters in TPOT are capped in a certain range. Thus, the mathematical operators will tend to produce values that are at the extremes of those caps, which is probably not ideal.

With these issues in mind, I propose that we remove the mathematical operators and have a fixed set of integer and float terminals that TPOT uses. We will need to choose these terminals wisely, as (if we make this change) TPOT will not be able to produce arbitrary terminals any more.

rhiever:
Considerations for `float` terminals:

* We will need good coverage in the range `[0., 1.]`, as many models and preprocessors use values in that range to specify percentages

* We will need good coverage in the much lower ranges (<= 0.01), as many models with the `C` and `learning_rate` parameters require lower float values

* We will need moderate coverage in the higher ranges (>=1), as some models benefit from e.g. higher `C` values. It's still unclear to us what specific values are useful here.

* SelectFwe uses `float`s in the range `[0.001, 0.05]`

* VarianceThreshold, Binarizer, RBFSampler, & Nystroem use `float`s in an unknown range


Considerations for `int` terminals:

* We use `int`s to encode multi-choice model parameters (e.g., options A, B, or C), so we need good coverage between 1-10

* The KNeighborsClassifier uses `int`s for its `n_neighbors` parameter, but `n_neighbors` doesn't seem to be useful beyond 25

* RFE & SelectKBest use `int`s to specify the number of features to select

* RandomizedPCA, RBF, FastICA, FeatureAgglomeration, & Nystroem use `int`s to specify the number of components

So we need to decide on terminal sets that account for all these use cases without creating a huge search space.

rhiever:
So perhaps a good `float` terminal set would be:

```
[0.0001, 0.001, 0.01] + np.arange(0.05, 1., 0.05) + np.arange(1, 50, 1)
```

[...]

PR #169, opened 2016-06-16
Remove math operators

teaearlgraycold:
## What does this PR do?

Removes `mul`, `add`, `sub`, and `_div` Primitives from TPOT

I discovered that after I removed the mathematical operators from TPOT that DEAP would get very upset during population generation because no Primitives generating `int`s and `float`s existed. DEAP will randomly select arguments from a Primitive and decide that those arguments need to be the spawn of new nodes in the tree. DEAP decides beforehand how big each tree should be, so if you only have a RF and you've already ignored the possibility of feeding a preprocessor into the RF then you *need* to now use some kind of `int` Primitive.

This is likely why math operators became so prevalent in TPOT. I've added a dumm() function as a `float`/`int` Primitive, so this branch can actually run, but that doesn't actually solve the problem. Now pipelines like like this:

`_knnc(_gaussian_nb(_extra_trees(ARG0, 32, 40)), dummy(dummy(31)))`

Relevant DEAP code: https://github.com/DEAP/deap/blob/master/deap/gp.py#L601

#162


tpot/tpot.py
```diff
@@ -169,17 +169,27 @@ def __init__(self, population_size=100, generations=100,
...
+
+        for val in float_terminals:
             self._pset.addTerminal(val, float)
 
+        # Remove this immediately
+        def dummy(foo):
+            return foo
+
+        self._pset.addPrimitive(dummy, [int], int)
+        self._pset.addPrimitive(dummy, [float], float)
```
rhiever: What is this? Did you mean to leave it in?

I think we were also going to add boolean terminals, right?

tpot/tpot.py
```diff
@@ -169,17 +169,27 @@ def __init__(self, population_size=100, generations=100,
...
 
-        for val in range(0, 101):
+        for val in int_terminals:
             self._pset.addTerminal(val, int)
-        for val in [100.0, 10.0, 1.0, 0.1, 0.01, 0.001, 0.0001]:
+
+        float_terminals = np.concatenate(([0., 1e-6, 1e-5, 1e-4, 1e-3],
+                np.linspace(0.01, 1.00, 100),
+                np.linspace(2., 50., 49),
+                np.linspace(60., 100., 5)))
```
rhiever: Let's rework this to:

np.concatenate(([1e-6, 1e-5, 1e-4, 1e-3, 1e-2, 1e-1],
                np.linspace(0.0, 1.00, 101),
                np.linspace(2., 50., 49),
                np.linspace(60., 100., 5)))

tpot/tpot.py
teaearlgraycold: I talked about that in the "background context" section of my PR.

The way DEAP makes a tree I need those Primitives in there, but they should absolutely not be necessary. Was wondering if you had any insight into preventing their necessity. Without them DEAP throws an exception because it might want to expand on an `int` parameter but can't do that if there's only `int` Terminals and not any `int` Primitives.

tpot/tpot.py
teaearlgraycold: Seems fine to me

tpot/tpot.py
rhiever: Oops, missed that. That really stinks. I do remember debugging something related to this before...

Do the `dummy` primitives count toward the pipeline size in the fitness evaluation? That could be one way to cut down on them, but at the same time, it would unfairly penalize operators with several parameters. Hrm...

rhiever:
Also just noticed that there's merge conflicts now.

tpot/tpot.py
teaearlgraycold: Honestly I feel that the `generate()` function of DEAP should be a bit more intelligent and exhaust all parameters before calling it quits and saying it can't build the tree any further.

rhiever: I agree. Maybe we should file a bug in DEAP? I'll post something on their mailing list.

teaearlgraycold: I think we may be able to fix the issue by reorganizing all parameters so that `input_df` comes last for all operators.

rhiever: Can you give that a try and see if it works? That seems odd, but... maybe!

teaearlgraycold: Tried it, and it seems I was wrong.

tpot/export_utils.py
```diff
@@ -388,7 +348,7 @@ def replace_function_calls(pipeline_list):
 
         elif operator_name == '_multinomial_nb':
             alpha = float(operator[3])
-            fit_prior = (int(operator[4]) % 2) == 0
+            fit_prior = int(operator[4])
```
rhiever: This needs to be `bool`, not `int`, I believe.

tpot/tpot.py
rhiever: Rework this part real quick please. :-)

tpot/export_utils.py
teaearlgraycold: Given what we've seen today that probably would have worked, but a good catch nonetheless.

Issue #154, opened 2016-05-23
Make XGBoost an optional dependency


rhiever:
Due to several users having install issues with XGBoost, we have decided to make XGBoost an optional dependency. Currently, XGBoost is not supported by TPOT. If you would like to implement XGBoost as an optional model in TPOT, please make suggestions here.

KhaoticMind:
I'll try to implement it and send a PR when I get it done.

This will be my first contribution to TPOT (and one of the few in general) so be patient ;)

rhiever:
Sounds good! Thanks @KhaoticMind. :-)

rhiever:
How's this coming along, @KhaoticMind? Need any feedback?

KhaoticMind:
Hello @rhiever. The last couple of weeks I'm having a lot of work at Day Job (where I can do most of my contribution). I hope to get something  done by next week or so.

In the mean time, If there is any one else interested on it, please don't be shy to jump in :) (But I'm still committed to it!).

rhiever:
No problem! We'll slate it for a later release than the one coming up this
week.

On Monday, June 20, 2016, Antonio Augusto Santos <notifications@github.com>
wrote:

> Hello @rhiever <https://github.com/rhiever>. The last couple of weeks I'm
> having a lot of work at Day Job (where I can do most of my contribution). I
> hope to get something done by next week or so.
>
> In the mean time, If there is any one else interested on it, please don't
> be shy to jump in :) (But I'm still committed to it!).
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/rhiever/tpot/issues/154#issuecomment-227115420>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/ABo7tzdR1AR3-wssVNI12chSNJ_zyALSks5qNnbigaJpZM4Ikwbk>
> .
>


-- 
Randal S. Olson, Ph.D.
Postdoctoral Researcher, Institute for Biomedical Informatics
University of Pennsylvania

E-mail: rso@randalolson.com | Twitter: @randal_olson
<https://twitter.com/randal_olson>
http://www.randalolson.com

KhaoticMind:
@rhiever Found some time, and already sent a pull request. Please be free to point any issues (as I said, my first big PR ;) .)

PR #183, opened 2016-06-21
Bring XGBoost back as an optional dependency

KhaoticMind:
## What does this PR do?
The objective of this PR is to put XGBoost back in TPOT as an Optional dependency, as discussed on issue #154 

I've brought the documentation back from prior commits and added:
- checks to see if xgboost is present
- tests to new xgboost parameters
- ability to export xgboost operator that are part of a pipeline

All nose tests (including new ones) are ok on my machine.
Tried running a toy test with just _xgb_boosting as a primitive in TPOT._pset and all was fine.

```
from tpot import TPOT
from sklearn.cross_validation import train_test_split

digits = load_digits()
                                                    train_size=0.75, test_size=0.25)

tpot = TPOT(generations=3, verbosity=2, population_size=5)
tpot.fit(X_train, y_train)
tpot.export('tpot_mnist_pipeline.py')
```

Not really...

Issue in #154

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -45,4 +45,5 @@ This project is hosted at https://github.com/rhiever/tpot
         'Topic :: Scientific/Engineering :: Artificial Intelligence'
     ],
     keywords=['pipeline optimization', 'hyperparameter optimization', 'data science', 'machine learning', 'genetic programming', 'evolutionary computation'],
+    extra_require = {'xgboost': ['xgboost']},
 )
--- a/tpot/export_utils.py
+++ b/tpot/export_utils.py
@@ -396,20 +396,22 @@ gbc{OPERATOR_NUM}.fit({OUTPUT_DF}.loc[training_indices].drop('class', axis=1).va
 """.format(OUTPUT_DF=result_name, OPERATOR_NUM=operator_num, LEARNING_RATE=learning_rate, MAX_FEATURES=max_features, MIN_WEIGHT=min_weight)
         
         elif operator_name == '_xg_boosting':
+            
             learning_rate = min(1., max(float(operator[3]), 0.0001))
-            subsample = min(1., max(0., float(operator[4])))
-            min_child_weight = min(0.5, max(0., float(operator[5])))
+            subsample = min(1., max(0.1, float(operator[4])))            
+            min_child_weight = min(100, max(0, int(operator[5])))
+            max_depth = min(100, max(1, int(operator[6])))            
 
             if result_name != operator[2]:
                 operator_text += "\n{OUTPUT_DF} = {INPUT_DF}.copy()".format(OUTPUT_DF=result_name, INPUT_DF=operator[2])
 
             operator_text += """
 # Perform classification with a extreme gradient boosting classifier
-xgbc{OPERATOR_NUM} = XGBClassifier(learning_rate={LEARNING_RATE}, subsample={SUBSAMPLE}, min_child_weight={MIN_WEIGHT}, n_estimators=500)
+xgbc{OPERATOR_NUM} = XGBClassifier(learning_rate={LEARNING_RATE}, subsample={SUBSAMPLE}, min_child_weight={MIN_WEIGHT}, max_depth={MAX_DEPTH}, n_estimators=500)
 xgbc{OPERATOR_NUM}.fit({OUTPUT_DF}.loc[training_indices].drop('class', axis=1).values, {OUTPUT_DF}.loc[training_indices, 'class'].values)
 
 {OUTPUT_DF}['xgbc{OPERATOR_NUM}-classification'] = xgbc{OPERATOR_NUM}.predict({OUTPUT_DF}.drop('class', axis=1).values)
-""".format(OUTPUT_DF=result_name, OPERATOR_NUM=operator_num, LEARNING_RATE=learning_rate, SUBSAMPLE=subsample, MIN_WEIGHT=min_child_weight)
+""".format(OUTPUT_DF=result_name, OPERATOR_NUM=operator_num, LEARNING_RATE=learning_rate, SUBSAMPLE=subsample, MIN_WEIGHT=min_child_weight, MAX_DEPTH=max_depth)
 
         elif operator_name == '_combine_dfs':
             operator_text += '\n# Combine two DataFrames'
--- a/tpot/export_utils.py
+++ b/tpot/export_utils.py
@@ -111,6 +111,7 @@ def generate_import_code(pipeline_list):
         '_ada_boost':           {'sklearn.ensemble': ['AdaBoostClassifier']},
         '_extra_trees':         {'sklearn.ensemble': ['ExtraTreesClassifier']},
         '_gradient_boosting':   {'sklearn.ensemble': ['GradientBoostingClassifier']},
+        '_xg_boosting':         {'xgboost': ['XGBClassifier']},
         '_logistic_regression': {'sklearn.linear_model': ['LogisticRegression']},
         '_passive_aggressive':  {'sklearn.linear_model': ['PassiveAggressiveClassifier']},
         '_linear_svc':          {'sklearn.svm': ['LinearSVC']},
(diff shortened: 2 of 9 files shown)


teaearlgraycold:
I'd suggest doing this instead:

And then in `__init__()`

```Python
try:
    from xgboost import XGBClassifier
    self._pset.addPrimitive(self._xg_boosting, [pd.DataFrame, float, float, int, int], pd.DataFrame)
except ImportError:
    self._pset.addPrimitive(self._gradient_boosting, [pd.DataFrame, float, float, float], pd.DataFrame)
```

Parameters for `_xg_boosting` might need tweaking, too.

KhaoticMind:
Hi @teaearlgraycold. 
1) I kept _gradient_boosting and _xg_boosting because I was under the impression that this was the intent. But this is easy to overcome. Just say the word.
2) I prefer to keep the import at the top of the file, and use a flag do detect error, to respect pep8
3) I'll update the setup.py to add an extra_requires flag, so to make xgboost explicitly optional.
4) I'll update the travis scripts to depend on XGBoost

Please comment on 1 and 2.
I'll push new commits for 3 and 4 soon.

setup.py
```diff
@@ -45,4 +45,5 @@ def calculate_version():
         'Topic :: Scientific/Engineering :: Artificial Intelligence'
     ],
     keywords=['pipeline optimization', 'hyperparameter optimization', 'data science', 'machine learning', 'genetic programming', 'evolutionary computation'],
+    extra_require = {'xgboost': ['xgboost']},
```
rhiever: I believe it's `extras_require`: http://peak.telecommunity.com/DevCenter/setuptools#declaring-extras-optional-features-with-their-own-dependencies

tpot/export_utils.py
```diff
@@ -393,6 +394,24 @@ def replace_function_calls(pipeline_list):
 
 {OUTPUT_DF}['gbc{OPERATOR_NUM}-classification'] = gbc{OPERATOR_NUM}.predict({OUTPUT_DF}.drop('class', axis=1).values)
 """.format(OUTPUT_DF=result_name, OPERATOR_NUM=operator_num, LEARNING_RATE=learning_rate, MAX_FEATURES=max_features, MIN_WEIGHT=min_weight)
+        
+        elif operator_name == '_xg_boosting':
+            
+            learning_rate = min(1., max(float(operator[3]), 0.0001))
+            subsample = min(1., max(0.1, float(operator[4])))            
+            min_child_weight = min(100, max(0, int(operator[5])))
+            max_depth = min(100, max(1, int(operator[6])))
```
rhiever: How did you decide these parameters? Do you know these parameters have a significant effect, and are capped in the right range? (Just curious, as I'm not terribly familiar with XGBoost in practice.)

tpot/export_utils.py
```diff
@@ -393,6 +394,24 @@ def replace_function_calls(pipeline_list):
...
+            min_child_weight = min(100, max(0, int(operator[5])))
+            max_depth = min(100, max(1, int(operator[6])))            
+
+            if result_name != operator[2]:
+                operator_text += "\n{OUTPUT_DF} = {INPUT_DF}.copy()".format(OUTPUT_DF=result_name, INPUT_DF=operator[2])
+
+            operator_text += """
+# Perform classification with a extreme gradient boosting classifier
+xgbc{OPERATOR_NUM} = XGBClassifier(learning_rate={LEARNING_RATE}, subsample={SUBSAMPLE}, min_child_weight={MIN_WEIGHT}, max_depth={MAX_DEPTH}, n_estimators=500)
+xgbc{OPERATOR_NUM}.fit({OUTPUT_DF}.loc[training_indices].drop('class', axis=1).values, {OUTPUT_DF}.loc[training_indices, 'class'].values)
```
rhiever: Is there a tradeoff between `n_estimators` and e.g. `learning_rate` in a XGBClassifer? Or is it a good assumption that `n_estimators=500` is a good setting?

KhaoticMind: They are closely related. A bigger learning rate usually requires a lower n_estimators to prevent overfitting.
For the xgboost Python interface the default values are learning_rate=0.1 and n_estimators=100 (http://xgboost.readthedocs.io/en/latest/python/python_api.html#module-xgboost.sklearn).

I limited n_estimators here just as it has been done with GrandientBoostingClassifier (where learning_rate / n_estimators play a similar (equal?) role).

It might be a good idea to let DEAP search for n_estimators as well as learning_rate. If this is the case, it should also be done for the GrandientBoostingClassifier.

tpot/export_utils.py
KhaoticMind: I've used learning_rate, subsample and min_child_weight as these are the counterparts of     learning_rate, max_features and min_weight present in _gradient_boosting. Also, from my experience, max_depth plays a very important role to prevent overfitting (https://github.com/dmlc/xgboost/blob/master/doc/how_to/param_tuning.md).

I got the value caps from https://github.com/dmlc/xgboost/blob/master/doc/parameter.md (learning_rate (eta) = [0,1], subsample = (0, 1], min_child_weight = [0, inifity], max_depth = [1, inifity]).
Base on these I used some common sense for:
-  subsample: 0 samples per tree makes no sense for me
- max_depth : a depth of 0 means no tree (for what I understand), and a tree that is too deep will surely overfit

setup.py
KhaoticMind: You are right. Fixed it on 2910acd.

tpot/export_utils.py
rhiever: IIRC GradientBoostingClassifier wasn't tremendously affected by `n_estimators` in our sklearn benchmarks. Maybe we need to look at that again.

tpot/export_utils.py
rhiever: Makes sense. If we have `min_child_weight`, then there is no need for `max_depth`. They both do the same thing (limit tree depth), albeit through different mechanisms.

In cases where a number is specified (e.g. `min_child_weight` seems to be a count?), is it possible to pass a float indicating a fraction instead?

KhaoticMind: Yeah. Haven't thought of that. We can keep just `min_child_weight` and make `max_depth` a really large  value (or `None` if this is accept by `xgboost`) to see if this works. I'll work on it.

For `min_child_weight`, (for what I know) it can't be a fraction meaning a fraction of the data. On `xgboost` the classes can have weights (created on the `DMatrix` datatype used internally by it), so it accepts fractions here just to reflect that some classes might have fractional weight.

In the case of the `sklearn` interface (`XGBClassifier`) all classes should have weight 1. So the min value here should be 1. I'll also fix this.

tpot/export_utils.py
KhaoticMind: Take a look here: http://pt.slideshare.net/DataRobot/gradient-boosted-regression-trees-in-scikitlearn
The parameter tuning section starts on slide 22. The effect of `learning_rate` and `n_estimators` can be seen on slide 25.

KhaoticMind:
Some more commits to fix some pending issues discussed above.
Please feel free to comment on this new version.

rhiever:
Hi @KhaoticMind, thank you so much for working XGBoost back into TPOT as an optional dependency! As you may have seen, we've been refactoring TPOT so it's easier to maintain. You can find the latest version of the refactor in the [development branch](https://github.com/rhiever/tpot/tree/development).

This change will require some rework of your PR. Would you prefer to rework the PR into the new framework, or would you like @teaearlgraycold and I to take your code and rework it?

[Also note: all PRs must be sent to the `development` branch now, as the `master` branch is intended for the latest stable version of TPOT.]

KhaoticMind:
Hello @rhiever! I think I'll prefer to leave it to @teaearlgraycold , as I'm currently short on time :(
I'll keep an eye on the project, and as soon as I have more spare time I'll head back and try to make some more contributions.

rhiever:
No problem! Thank you for your contributions. :-)

Issue #156, opened 2016-05-29
Rework custom scoring functionality


mglowacki100:
Now, as far I understood there is only "balanced class accuracy" as default;  precision, recall and f1 are hard-coded in: http://rhiever.github.io/tpot/examples/Custom_Scoring_Functions/
Do you consider to add other metrics either as hard-coded custom scoring function or wrappers on:
http://scikit-learn.org/stable/modules/classes.html#module-sklearn.metrics ? I'm especially interested in log_loss metric.

rhiever:
As of now, you would have to write a wrapper function as shown in the docs page you listed. Now that I look at that, though, that doesn't seem very ideal. I'm going to turn this issue into an enhancement request to improve the custom scoring functionality. It should be possible to pass an arbitrary scoring function from sklearn and allow TPOT to use that, without writing a custom wrapper function.

bartleyn:
I can whip something up for this. I think there was an initial concern about how to optimize the scoring functions for DataFrames, but it should be pretty straightforward to just ship off all those concerns to the underlying sklearn functions.

rhiever:
Thanks @bartleyn!

mglowacki100:
Great !!!

rhiever:
Just wanted to ping you, @bartleyn, to see how this was coming along.

mglowacki100:
@bartleyn when do you plan to release this functionality?

bartleyn:
Sorry, I've been horribly negligent of my duties on this (I had a personal emergency to deal with). I've got the lion's share of it done, but need to test edge cases (multiclass, multilabel, etc) a little more thoroughly before putting up a PR. I should comfortably have something by the weekend.

rhiever:
This change will be in the upcoming 0.5 release.

PR #199, opened 2016-07-18
[WIP] Custom scoring

bartleyn:
## What does this PR do?
 Addresses #156 to allow for classification metrics from scikit-learn's metrics module to be passed to TPOT as scoring functions, along with keyword arguments for the function. Currently  lumps together scoring functions that require decision functions and prediction probabilities together. 

- [x] actual implementation 
- [x] unit-tests for different metrics
- [x] documentation change

fit(), _evaluate_individual(), and _train_model_and_predict()

With classification scoring functions from sklearn.metrics

Some of the functions require a classifier's _decision-function_ output in place of a list of predicted classes, and some of the functions require a classifier's _prediction-probabilities_ output. These functions are currently lumped together in consideration -- it is difficult to inspect the input scoring function to see if it requires one or the other (without explicit listing of all the scoring functions, which would become bulky when we start considering regression metrics, etc), so scoring functions requiring decision functions _may_ mis-represent an individual's fitness.  

It was decided to lump decision function scoring functions into the prediction probability ones (and not vice versa), since decision functions can return negative numbers, and probabilities not. 

#156 

## Questions:

Yes, at the very least to reflect the caveats discussed above about the decision_function. 
No. 
- How much should TPOT handle wrt type checking, etc, and how much should we rely on the users?

### DIFF
--- a/tpot/operators/classifiers/base.py
+++ b/tpot/operators/classifiers/base.py
@@ -32,7 +32,6 @@ class Classifier(Operator):
     def _call(self, input_matrix, *args, **kwargs):
         # Calculate arguments to be passed directly to sklearn
         operator_args = self.preprocess_args(*args, **kwargs)
-
         return self._train_and_predict(input_matrix, operator_args)
 
     def _train_and_predict(self, input_matrix, operator_args):
@@ -57,8 +56,22 @@ class Classifier(Operator):
         clf.fit(self.training_features, self.training_classes)
 
         all_features = np.delete(input_matrix, non_feature_columns, axis=1)
-        input_matrix[:, GUESS_COL] = clf.predict(all_features)
+        if self.clf_eval_func == 'decision_function':
+            if 'decision_function' in dir(clf):
+                input_matrix = np.insert(input_matrix, GUESS_COL, clf.decision_function(all_features).T, axis=1)
+            else:
+                input_matrix = np.insert(input_matrix, GUESS_COL, np.array([[0.] * max(1, len(np.unique(training_classes))) for x in list(range(input_matrix.shape[0]))]).T, axis=1)
+        elif self.clf_eval_func == 'predict_proba':
+            if 'predict_proba' in dir(clf):
+                input_matrix = np.insert(input_matrix, GUESS_COL, clf.predict_proba(all_features).T, axis=1)
+            else:
+                input_matrix = np.insert(input_matrix, GUESS_COL, np.array([[0.] * max(1, len(np.unique(training_classes))) for x in list(range(input_matrix.shape[0]))]).T, axis=1)
+        else:
+            input_matrix[:, GUESS_COL] = clf.predict(all_features)
         # Store the guesses as a synthetic feature
-        np.insert(input_matrix, input_matrix.shape[1], input_matrix[:, GUESS_COL], axis=1)
+        if self.clf_eval_func == 'predict_proba' or self.clf_eval_func == 'decision_function':
+            input_matrix = np.insert(input_matrix, input_matrix.shape[1], clf.predict(all_features), axis=1)
+        else:
+            input_matrix = np.insert(input_matrix, input_matrix.shape[1], input_matrix[:, GUESS_COL], axis=1)
 
         return input_matrix
--- a/tpot/operators/classifiers/base.py
+++ b/tpot/operators/classifiers/base.py
@@ -56,22 +56,21 @@ class Classifier(Operator):
         clf.fit(self.training_features, self.training_classes)
 
         all_features = np.delete(input_matrix, non_feature_columns, axis=1)
+
+        input_matrix[:, GUESS_COL] = clf.predict(all_features)
+
+        # Store the guesses as a synthetic feature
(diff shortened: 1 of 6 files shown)


bartleyn:
After digging around for some code-inspection ideas, I found that you can access the docstring of the sklearn metrics with commands like `help()`. With some straightforward parsing, this would allow for flexibility when it comes to using a classifier's `decision_function` versus `predict_proba` (because the relevant sklearn metric docstrings specify which function to use), and eliminates the need to warn the user about this issue. I'll update accordingly.

rhiever:
Hi @bartleyn, thank you so much for reworking the scoring system! As you may have seen, we've been refactoring TPOT so it's easier to maintain. You can find the latest version of the refactor in the [development branch](https://github.com/rhiever/tpot/tree/development).

This change will require some slight rework of your PR. Would you prefer to rework the PR into the new framework, or would you like @teaearlgraycold and I to take your code and rework it?

bartleyn:
I've pulled those changes -- the refactor looks great. I should at least be able to resolve the merge conflicts in the main tpot.py (and maybe the tests). But maybe I'll hand it off at that point, since I'm not quite sure (yet) where things plug into the new framework.

rhiever:
@teaearlgraycold can help with any complications arising from or explanations of the refactor. :-)

tpot/tpot.py
```diff
@@ -485,314 +468,6 @@ def _parse_scoring_docstring(self, scoring_function):
...
-        """
-        learning_rate = min(1., max(learning_rate, 0.0001))
-        max_features = min(1., max(0., learning_rate))
-        min_weight = min(0.5, max(0., min_weight))
-
-        return self._train_model_and_predict(input_df, GradientBoostingClassifier,
-            learning_rate=learning_rate, n_estimators=500,
-            max_features=max_features, random_state=42, min_weight_fraction_leaf=min_weight)
-
     def _train_model_and_predict(self, input_df, model, **kwargs):
```
bartleyn: @teaearlgraycold  I resolved most conflicts (and tried to adopt the new numpy matrices/indices), but I left my original _train_model_and_predict() in because I wanted to see what you think the best way is for passing the clf_eval_func var to the separate classifier operators. In writing this I think I figured it out though -- throw it into a Classifier's _call and keep track of it there.

tpot/operators/classifiers/base.py
```diff
@@ -57,8 +56,22 @@ def _train_and_predict(self, input_matrix, operator_args):
         clf.fit(self.training_features, self.training_classes)
 
         all_features = np.delete(input_matrix, non_feature_columns, axis=1)
-        input_matrix[:, GUESS_COL] = clf.predict(all_features)
+        if self.clf_eval_func == 'decision_function':
+            if 'decision_function' in dir(clf):
+                input_matrix = np.insert(input_matrix, GUESS_COL, clf.decision_function(all_features).T, axis=1)
```
bartleyn: @teaearlgraycold should we be updating the non_feature_columns in the classifiers that add multiple columns in at GUESS_COL?

tpot/tpot.py
teaearlgraycold: Yeah that sounds good

tpot/operators/classifiers/base.py
teaearlgraycold: >classifiers that add multiple columns in at `GUESS_COL`

I don't believe a classifier should be adding multiple columns in where `GUESS_COL` is.

bartleyn: Sorry about that. To clarify: certain scoring functions need an array of labels of size (n_samples,) as well as a matrix of prediction probabilities (or in some cases, decision function values) of size (n_samples, n_classes). This means using either the predict_proba() or decision_function() methods (and putting filler in if the classifier doesn't implement the function necessary). Originally I had been pickling these values into strings and putting them in the single guess column, but that was assuming an underlying dataframe. Rather than gutting the existing numpy matrix code, I figured it'd be easier to insert the (n_samples, n_classes) into the matrix at GUESS_COL and just try to keep track of all the added columns.

teaearlgraycold: If you need to store more data you'll probably be best off adding another "non_feature_column" entirely - not just keeping a reference to a column after `GUESS_COL`. I never really considered that contingency, though, so that might break all other code dealing with column indices.

rhiever: For now, I think we should only support outputs and scorers that output only a single value. Multi-label classification may eventually require a set up that supports multiple outputs, but that's a fair bit down the line.

bartleyn: You all helped me realize that I was totally overthinking it/ignoring the bigger picture. I think a solution to make this work is to still insert the classifier's `predict()` outputs in GUESS_COL, but to tack on anything regarding `predict_proba()` or `decision_function()` to the end of the matrix, and to use that slice at the end in the scoring functions as necessary. I've updated accordingly.

rhiever:
@bartleyn: Just merged this PR and @teaearlgraycold will jump in to wrap it up for the 0.5 release (hopefully this week). Thank you again!

PR #217, opened 2016-08-19
Evolve directly on sklearn pipelines

teaearlgraycold:
## What does this PR do?

Redesigns how TPOT evaluates pipelines, by changing the DEAP compile function to use the TPOT export utils' generate_pipeline_code function.

The changes also seem to fix the issue the development branch has had with scoring pipelines.

Notable changes/additions:

* All `_call` and `__call__` code removed from the operators (`CombineDFs` operator is now a shell of its former self)
* `_evaluate_individual`
* `_compile_to_sklearn`

Tests are not yet updated for these changes, but it would be a good idea to remove the "catch-all" exception handler from `_evaluate_individual` and see what mistakes are happening under the hood.


teaearlgraycold:
Given how much code this removes from TPOT, I'm glad to say I was wrong about what I said about this change. It's a great improvement upon the over-complexity of TPOT.

rhiever:
Testing this change now. It looks elegant. I love it!

Some initial notes:

1) A bunch of unit tests are broken because of all the removed code

2) Balanced accuracy should be the default scoring function, not regular accuracy. Where did our balanced accuracy function go?

3) We should explicitly specify `cv=3` in `cross_val_score`

4) I keep getting this `UserWarning`:

```
anaconda/lib/python3.5/site-packages/sklearn/feature_selection/base.py:80: UserWarning: No features were selected: either the data is too noisy or the selection test too strict.)
```

5) In all cases throughout TPOT, set `n_jobs=1` instead of `-1`. We should not assume that the user would like to use multithreading. By default, all sklearn `n_job` values are `1`, so we just need to remove all cases in TPOT that set it to `-1`.

rhiever:
I'm just gonna merge this and start hacking at some of these issues. Can you work on the unit tests in the meantime?

teaearlgraycold:
Sure

PR #222, opened 2016-08-19
Tpot mdr#2

sohnam:

## Questions:


tpot/tpot.py
```diff
@@ -73,7 +72,8 @@ class TPOT(object):
     def __init__(self, population_size=100, generations=100,
                  mutation_rate=0.9, crossover_rate=0.05,
                  random_state=0, verbosity=0, scoring_function=None,
-                 disable_update_check=False):
+                 disable_update_check=False, expert_source=None,
+                 ekf_index=0, k_best=5):
```
rhiever: No need to set ekf_index and k_best in the TPOT init method.

tpot/tpot.py
```diff
@@ -247,6 +246,10 @@ def fit(self, features, classes):
         None
 
         """
+        
+        if self.expert_source != None:
```
rhiever: Use `is not None` instead of `!= None` to make sure the check succeeds.

rhiever: Actually, I'm guessing this was debug code that can be removed. :-) But the point still stands when checking for `None` values.

tpot/tpot.py
```diff
@@ -1068,6 +1072,39 @@ def _select_fwe(self, input_df, alpha):
...
+#        all_ekf_cols = len(training_features.columns)
+        training_features = input_df
+        all_ekf_cols = training_features.shape[1]
+    
+        if k_best < 1:
+            k_best = 1
+        elif k_best >= all_ekf_cols:
+            k_best = all_ekf_cols        
+        
+        if self.expert_source[ekf_index].dtype==int:
```
rhiever: Make this check something like `set(self.expert_source[ekf_index]) == set([True, False])` to guarantee that we're using a mask here.

tpot/tpot.py
```diff
@@ -1068,6 +1072,39 @@ def _select_fwe(self, input_df, alpha):
...
+            ekf_source = self.expert_source[ekf_index].astype(bool)
+            return training_features[:, ekf_source].copy()
+        else:
+            ekf_source = np.argsort(self.expert_source)[::-1][:10]
```
rhiever: Should the `[:10]` be there? Isn't that limiting the result to only the top 10 indices?

tpot/tpot.py
```diff
@@ -1068,6 +1072,39 @@ def _select_fwe(self, input_df, alpha):
...
+        
+        Returns
+        -------
+        subsetted_df: andas.DataFrame {n_samples, n_filtered_features + ['guess', 'group', 'class']}
+            Returns a DataFrame containing the 'best' features provided by expert_source
+        
+        """
+#        training_features = input_df.loc[input_df['group'] == 'training'].drop(self.non_feature_columns, axis=1)
+#        all_ekf_cols = len(training_features.columns)
+        training_features = input_df
```
rhiever: No need to re-assign this variable: we can use the `input_df` variable directly.

tpot/tpot.py
```diff
@@ -247,6 +246,10 @@ def fit(self, features, classes):
         None
 
         """
+        
+        if self.expert_source != None:
+            self._ekf(features, self.ekf_index, self.k_best)        
+
```
rhiever: We should do some checks on the `expert_source` values at `fit` time. In particular:

1) Make sure all EK sources are of the same length as the input features

2) Make sure all EK sources with only the values `0` and `1` are encoded as `bool`

3) Make sure there's no missing values in the EK sources

rhiever:
Please let me know when you think this is ready for review.

sohnam:
Uploaded the latest changes (checks).

rhiever:
Looks like we have some merge conflicts -- can you pull from this `tpot-mdr` branch and sort out the conflict?

sohnam:
I pulled from the main tpot-mdr branch, added the ekf changes and re-uploaded the file.

PR #230, opened 2016-08-20
Add SelectFromModel feature selector (Fixes #212)

### LINKED ISSUE #212: Add new feature selection operator based on ExtraTrees feature importances
Use [SelectFromModel](http://scikit-learn.org/stable/modules/generated/sklearn.feature_selection.SelectFromModel.html) to select features based on [ExtraTrees](http://scikit-learn.org/stable/modules/generated/sklearn.ensemble.ExtraTreesClassifier.html) feature importances.

`threshold` should be an evolvable `float` parameter.

For the ExtraTrees model, we should open the same parameters to evolution as we do for the [ExtraTreesClassifier operator](https://github.com/rhiever/tpot/blob/development/tpot/operators/classifiers/extra_trees.py#L46).

teaearlgraycold:
## What does this PR do?

* Adds the SelectFromModel feature selector, using ExtraTreesClassifiers' feature importances

I'm not sure exactly what min and max should be set on threshold for SelectFromModel. Right now I have it set to be in the range of 0. to 1.0

`tpot/operators/selectors/select_from_model.py`

#212

### DIFF
--- a/tpot/tpot.py
+++ b/tpot/tpot.py
@@ -457,7 +457,7 @@ class TPOT(object):
         """
 
         try:
-            # Transform the tree expression in a callable function
+            # Transform the tree expression into an sklearn pipeline
             sklearn_pipeline = self._toolbox.compile(expr=individual)
 
             # Count the number of pipeline operators as a measure of pipeline
@@ -481,6 +481,8 @@ class TPOT(object):
         except Exception:
             # Catch-all: Do not allow one pipeline that crashes to cause TPOT
             # to crash. Instead, assign the crashing pipeline a poor fitness
+            # import traceback
+            # traceback.print_exc()
             return 5000., 0.
         finally:
             if not self.pbar.disable:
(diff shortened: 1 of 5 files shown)


tpot/tpot.py
```diff
@@ -481,6 +481,8 @@ def _evaluate_individual(self, individual, features, classes):
         except Exception:
             # Catch-all: Do not allow one pipeline that crashes to cause TPOT
             # to crash. Instead, assign the crashing pipeline a poor fitness
+            # import traceback
+            # traceback.print_exc()
```
rhiever: Remove these maybe?

rhiever:
I think we need a test using this new operator in a pipeline.

teaearlgraycold:
I've tested it in a pipeline and didn't get any exceptions.

I'll write a test that generates pipelines for each operator and tries to fit each one.

PR #241, opened 2016-08-25
Have TPOT inherit from sklearn's BaseEstimator

teaearlgraycold:
## What does this PR do?

Changes the TPOT class so that it inherits from sklearn's `BaseEstimator`.

In order to do this, I needed to move the `balanced_accuracy` scoring function (due to recursive calls to `__repr__`) outside of the TPOT class, but that actually seems like an appropriate move now that it's a properly formatted scoring function.

Running the tests, manually checking the output of `get_params` and `set_params`.

Appropriate tests have been added to the test suite


tpot/tpot.py
```diff
@@ -452,7 +443,8 @@ def _compile_to_sklearn(self, expr):
         return eval(sklearn_pipeline, self.operators_context)
 
     def _evaluate_individual(self, individual, features, classes):
-        """Determines the `individual`'s fitness according to its performance on the provided data
+        """Determines the `individual`'s fitness according to its performance on
+        the provided data
```
rhiever: Just a note: The initial short description in the docstrings should be on one line only.

teaearlgraycold: Oh, didn't know about that.

The battle to keep line lengths down continues

rhiever: I'm generally 👎 on enforcing the 79-char line limit. ~160 seems more reasonable for modern screens.

Issue #238, opened 2016-08-24
Unable to use logloss as score function


bear0330:
tpot 0.5 supports all sklearn score functions but if I use logloss as score function, I will get this error:
![image](https://cloud.githubusercontent.com/assets/9865603/17934076/26ed6130-6a49-11e6-9af4-2a62090b0838.png)

## Context of the issue

If I use logloss function like this:
```
tpot = TPOT(generations=5, scoring_function='log_loss', random_state=42, 
            verbosity=2)
```

I will get the error above. I think this is because the code inside tpot.py line 302~307:

```
               top_score = 0.
                for pipeline_num, pipeline in enumerate(self.hof.items):
                    if self.hof.keys[pipeline_num].wvalues[1] > top_score:
                        self._optimized_pipeline = pipeline
```

here the top_score is started with 0, but logloss will be negative value, that causes the condition won't hit then self._optimized_pipeline won't be assigned.

This is just my guess, I haven't test it yet, but if I don't use logloss, this error disappears.

## Process to reproduce the issue

1. User creates TPOT instance
2. User calls TPOT `fit()` function with training data and use logloss as score function.
3. TPOT crashes with a `TypeError`.

## Expected result

It should work with logloss function.

## Current result

Crash.

## Possible fix

See my guess above.

## `name of issue` screenshot

See introduction.

rhiever:
Checking: Are you using TPOT for a regression problem? TPOT doesn't currently support regression, but we hope to do so in the near future (0.6).

Thank you for reporting this issue! This made me realize that we should place a limit on what scoring functions that TPOT accepts at the moment, as log loss is not appropriate for classification problems.

rhiever:
Docs are updated with the latest push. Sorry for the confusion!

bartleyn:
@rhiever There is a change of mine that got lost in the 0.5.0 release that accounts for the loss functions, which I think might be (at least) part of the reason for this. I agree that there should maybe be limits on what scoring functions TPOT uses, but I can put this back in if you like.

rhiever:
Sounds good to me @bartleyn -- please roll it out as a patch to the dev branch.

PR #244, opened 2016-08-27
cross_val_score scoring_function support

bartleyn:
## What does this PR do?

1. Addresses #238, allowing use for compatible loss/error functions and having TPOT takes the best pipeline when using log_loss. 
2. Uses sklearn's get_scorer function to validate strings passed in a scoring_function are valid (for cross_val_score), and restricts use of custom functions. 
3. Tweaks the instantiator test to reflect the above changes. 

tpot.py:__init__()
tpot.py:fit()
tpot.py:_evaluate_individual()

Making sure that the fitted pipelines 

1. Some scoring functions implemented in sklearn only allow for binary classification, causing pipelines to raise exceptions if the data is multiclass.
2.  `log_loss` needs predict_proba support. This may restrict the classifier operators used in the final pipelines to those that support that functionality. 

#238 

## Questions:

No, as the docs already reflect that strings should be passed in as the scoring_function parameter.
No

### DIFF
--- a/docs/sources/using.md
+++ b/docs/sources/using.md
@@ -60,8 +60,8 @@ TPOT offers several arguments that can be provided at the command line:
 <tr>
 <td>-scoring</td>
 <td>SCORING_FN</td>
-<td>"accuracy", "adjusted_rand_score", "average_precision", "f1", "f1_macro", "f1_micro", "f1_samples", "f1_weighted", "precision", "precision_macro", "precision_micro", "precision_samples", "precision_weighted", "recall", "recall_macro", "recall_micro", "recall_samples", "recall_weighted", "roc_auc"</td>
-<td>Function used to evaluate the goodness of a given pipeline for the classification problem. By default, balanced class accuracy is used. TPOT assumes that this scoring function should be maximized, i.e., higher is better.</td>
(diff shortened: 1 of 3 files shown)


teaearlgraycold:
Why restrict scoring_function to strings? We already need to use functions as scoring functions due to balanced_accuracy.

bartleyn:
I made that decision because that was how I interpreted #238 and the decision to restrict possible functions to those supported by cross_val_score. I can revert that change (and will since it seems to break the build around fetching the params). 

I guess the question is, what do we allow then? We can go completely the way of `cross_val_score()` and allow both the appropriate strings and all callables with the signature `scorer(estimator, X, y)`. However, since we don't actually evaluate the scoring_function until we call `cross_val_score()` within `_evaluate_individual()`, any malformed scoring_functions will invalidate every pipeline (since we have that catchall Exception handler). What do you think would be elegant for this? Is a check for three args going to suffice?

teaearlgraycold:
We're already adding code to tell users what went wrong when all pipelines fail to evaluate.

As is standard with Python, you just have to trust the user will pass valid objects as parameters.

bartleyn:
Per your note, I've removed the input validation, and updated some of the info regarding scoring functions in the docs.

Issue #246, opened 2016-08-31
Add warm_start functionality


rhiever:
Some sklearn operators offer a `warm_start` functionality that, when set to `True`, reuses and builds upon the model constructed from the previous `fit` call on the object. We should offer the same functionality with TPOT.

Since TPOT has a population (`pop`) and a Pareto front (`self._hof`) at the end of the run, we can store these objects after the `fit` call and have the `fit` function reuse these objects if `fit` has previously been called.

`warm_start` should be `False` by default.

kamalasaurus:
Hi!  So I made a crack at this after v0.6 appeared to push up, here's what I've got so far:

https://github.com/kamalasaurus/tpot/commit/e8f47f997255250043946f55a445595bb54cdf49

I'm storing/not-rewriting pop and _hof in the case of a warm_start; I -think- that was the general plan.

I've only made additions to tpot/base.py .  The existing nosetests all seem to run fine, but I feel like I should add some others–I'm not 100% sure what to test in terms of the tpot lifecycle and when the stored variables should exist, so I'd appreciate some guidance on that.  Would it be easier to do in the form of a Pull Request?

Also, I looked at driver.py (is this feature also meant to be accessible from the command line?) and thought I'd add something like this:

```python
    parser.add_argument('--warm-start', action='store', dest='WARM_START', default=False,
        choices=[True, False], type=bool, help='use previously generated models')
```

But I'm not entirely sure if that makes any sense (can the command line pass boolean arguments?  Wouldn't the api be better being something like `-ws` which if added will flag `warm_start` to `True`?) so yeah, some guidance there would also be appreciated as well.  I'm not 100% clear on how tpot is called from the command line either...  does it boot up a new python process whenever you do so?  In which case, is warm_start even possible in that context?

So!  That's where I'm at, hopefully I didn't miss the boat completely!

Let me know what you think at your convenience!

PR #266, opened 2016-09-08
adding warm_start functionality by storing  pop and self._hof if desired

kamalasaurus:
## What does this PR do?

Hi!  So I made a crack at this after v0.6 appeared to push up, here's what I've got so far:

I'm storing/not-rewriting pop and _hof in the case of a warm_start; I -think- that was the general plan.

I've only made additions to tpot/base.py .  The existing nosetests all seem to run fine, but I feel like I should add some others–I'm not 100% sure what to test in terms of the tpot lifecycle and when the stored variables should exist, so I'd appreciate some guidance on that.  Would it be easier to do in the form of a Pull Request?

Also, I looked at driver.py (is this feature also meant to be accessible from the command line?) and thought I'd add something like this:

``` python
    parser.add_argument('--warm-start', action='store', dest='WARM_START', default=False,
        choices=[True, False], type=bool, help='use previously generated models')
```

But I'm not entirely sure if that makes any sense (can the command line pass boolean arguments?  Wouldn't the api be better being something like `-ws` which if added will flag `warm_start` to `True`?) so yeah, some guidance there would also be appreciated as well.  I'm not 100% clear on how tpot is called from the command line either...  does it boot up a new python process whenever you do so?  In which case, is warm_start even possible in that context?

So!  That's where I'm at, hopefully I didn't miss the boat completely!

Let me know what you think at your convenience!

Could use some guidance on that!
#246
## Questions:
  Yes, if the warm_start functionality is exposed to the end user.
  Nope.

### DIFF
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -71,7 +71,7 @@ class TPOTBase(BaseEstimator):
                  mutation_rate=0.9, crossover_rate=0.05,
                  scoring=None, num_cv_folds=3, max_time_mins=None, max_eval_time_mins=5,
                  random_state=None, verbosity=0,
-                 disable_update_check=False):
+                 disable_update_check=False, warm_start=False):
         """Sets up the genetic programming algorithm for pipeline optimization.
 
         Parameters
@@ -128,6 +128,9 @@ class TPOTBase(BaseEstimator):
             0 = none, 1 = minimal, 2 = all
         disable_update_check: bool (default: False)
             Flag indicating whether the TPOT version checker should be disabled.
+        warm_start: bool (default: False)
+            Flag indicating whether TPOT will reuse models from previous calls to
+            fit for faster operation
 
         Returns
         -------
@@ -146,6 +149,8 @@ class TPOTBase(BaseEstimator):
         self._hof = None
         self._optimized_pipeline = None
         self._fitted_pipeline = None
+        self._pop = None
+        self.warm_start = warm_start
         self.population_size = population_size
         self.generations = generations
         self.max_time_mins = max_time_mins
@@ -293,7 +298,8 @@ class TPOTBase(BaseEstimator):
         self._start_datetime = datetime.now()
 
         self._toolbox.register('evaluate', self._evaluate_individual, features=features, classes=classes)
-        pop = self._toolbox.population(n=self.population_size)
+        # assign population, self._pop can only be not None if warm_start is enabled
+        pop = self._pop if self._pop else self._toolbox.population(n=self.population_size)
 
         def pareto_eq(ind1, ind2):
             """Determines whether two individuals are equal on the Pareto front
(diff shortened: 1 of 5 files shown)


tpot/base.py
```diff
@@ -275,7 +280,8 @@ def fit(self, features, classes):
         self._start_datetime = datetime.now()
 
         self._toolbox.register('evaluate', self._evaluate_individual, features=features, classes=classes)
-        pop = self._toolbox.population(n=self.population_size)
+        # assign population, self._pop can only be not None if warm_start is enabled
+        pop = self._pop if self._pop else self._toolbox.population(n=self.population_size)
```
rhiever: I know it's less space-efficient, but can you please break this out into a multi-line if/then? Easier to read in the line run.

rhiever:
This code looks great! It's exactly what I had in mind for the issue. Thank you for coding it up.

>Also, I looked at driver.py (is this feature also meant to be accessible from the command line?) and thought I'd add something like this:

Warm start functionality unfortunately can't be made available to command-line users since the TPOT instance gets garbage collected as soon as the command-line call ends.

>The existing nosetests all seem to run fine, but I feel like I should add some others–I'm not 100% sure what to test in terms of the tpot lifecycle and when the stored variables should exist, so I'd appreciate some guidance on that.

I think we should make sure to update the unit tests at the very least. There's a test_init unit test that should check to make sure all variables assigned in init are properly assigned to a new TPOT instance (both default and user-provided params).

I think this PR should also add a new unit test that runs a short TPOTClassifier run (1 gen for 20 pop?) on the MNIST practice data set with `warm_start=True` and a fixed `random_state`, checks that the `_pop` and `hof` were assigned to something (e.g. `hof` should have at least one individual in it), runs another small run on that same TPOTClassifier instance (same data set, pop size, gens, different but fixed random state) and does the check again. That test will guarantee, at the very least, that the basic warm start functionality still works.

---

I also think we should consider renaming the hof variable throughout. Can you please change all references to `hof` to `pareto_front` throughout all files? (To make it easier: `hof` shouldn't be referenced in any of the operator files.)

Lastly, please make sure to update the docs (in this PR) with the new warm_start option.

kamalasaurus:
Okay, cool~

Thanks for the quick turnaround!  Sorry it took me a couple days to update, I was airplaned for a bit @_@

Anyway!  I think I've addressed all your comments, `hof` is now `pareto_front` in all mentions, docs are updated (I think), tests are added (I think) and the one-liner is now a multi-liner.

The test only has a population of 1, though (is that a facile test?) because when I set it to 20 the unit test became realllllly slow.  It might be good for you to do a basic sanity check (I'm not sure what's necessarily sane in this world!) to make sure everything's fine.  Please look over the test at tests.py:199 as well to make sure my test logic is correct as well.

Let me know what you think!

Issue #252, opened 2016-09-01
Refine TPOT parameters


rhiever:
This is an extension to #206. Here, we will define specific parameter ranges for each individual operator's parameter, so we don't have to apply caps to the parameters any more. Many of these parameters are from auto-sklearn, with modifications where we know more about the parameter space.
# Classifiers
## AdaBoost

learning_rate: [1e-3, 1e-2, 1e-1, 0.5, 1.]
algorithm: ["SAMME.R", "SAMME"]
max_depth: range(1, 11)
## BernoulliNB

alpha: [1e-3, 1e-2, 1e-1, 1., 10., 100.]
fit_prior: [True, False]
## DecisionTree

criterion: ["gini", "entropy"]
max_depth: range(1, 11)
min_samples_split: range(2, 21)
min_samples_leaf: range(1, 21)
## ExtraTrees

criterion: ["gini", "entropy"]
max_features: np.arange(0, 1.01, 0.05)
min_samples_split: range(2, 21)
min_samples_leaf: range(1, 21)
bootstrap: ["True", "False"]
## GradientBoosting

learning_rate: [1e-3, 1e-2, 1e-1, 0.5, 1.]
max_depth: range(1, 11)
min_samples_split: range(2, 21)
min_samples_leaf: range(1, 21)
subsample: np.arange(0.05, 1.01, 0.05)
max_features: np.arange(0, 1.01, 0.05)
## KNeighbors

n_neighbors: range(1, 101)
weights: ["uniform", "distance"]
p: [1, 2]
## LinearSVC

penalty: ["l1", "l2"]
loss: ["hinge", "squared_hinge"]
dual: [True, False]
tol: [1e-5, 1e-4, 1e-3, 1e-2, 1e-1]
C: [1e-4, 1e-3, 1e-2, 1e-1, 0.5, 1., 5., 10., 15., 20., 25.]

In preprocess, disallow:
- penalty = 'l1' and loss = 'hinge' (set loss = 'squared_hinge')
- dual = False and penalty = 'l2' and loss = 'hinge' (set loss = 'squared_hinge')
- dual = False and penalty = 'l1' (set dual = True)
## LogisticRegression

penalty: ["l1", "l2"]
C: [1e-4, 1e-3, 1e-2, 1e-1, 0.5, 1., 5., 10., 15., 20., 25.]
dual: [True, False]

In preprocess, disallow:
- dual = False and penalty = 'l1' (set dual = True)
## MultinomialNB

[...]

Issue #206, opened 2016-08-13
Refine TPOT's pipeline grammar


rhiever:
DELFT uses a relatively sophisticated [pipeline grammar](https://github.com/EpistasisLab/delft/blob/c924d61c885b13b4a82d86148e0b1fabd03a9cb0/delft/delft.py#L316) to restrict the forms that pipelines can take. For example:

1) An imputer must be the first step in the pipeline (see #194)

2) A scaler must be the second step in the pipeline

3) Only classifiers can be the final step in the pipeline

4) DELFT also has special terminal types for each parameter, such that it never, e.g., uses integers to encode string parameters (see #189)

Even if we keep the rest of the pipeline grammar relatively open for GP to tinker with, we should integrate all of these restrictions into TPOT's grammar.

rhiever:
@teaearlgraycold, how much of this have we already integrated into the dev branch?

teaearlgraycold:
So far -

- [ ] An imputer must be the first step in the pipeline (see #194)

- [ ] A scaler must be the second step in the pipeline

- [X] Only classifiers can be the final step in the pipeline

- [ ] special terminal types for each parameter, such that it never, e.g., uses integers to encode string parameters (see #189)

rhiever:
Great. We'll focus on this once the refactor is rolled out.

rhiever:
@teaearlgraycold, for 0.6 we're only implementing `3)` and `4)`, right?

PR #267, opened 2016-09-09
Refine parameters for estimators and create specialized parameter types  for GP

teaearlgraycold:
## What does this PR do?

Refines the parameter search space for TPOT operators, and defines specialized data types for each parameter type. Also bumps the version number to `0.7.0` since this is a `0.7` release feature.
- `tpot/operators/base.py`
- `tpot/operators/gp_types.py`

And in the specific operator modules.

Any logic added in this should be covered by the test suite.
#252
#206


teaearlgraycold:
I should say here that doing this actually breaks the `deap.Individual.from_string` function. I don't think it likes that there are multiple terminal sets that have the same terminal value (So, both `IteratedPower` and `MaxDepth` contain the number `2`).

Issue #253, opened 2016-09-01
Add Imputer operator


rhiever:
This issue branches off from #206.

TPOT should detect whether there is missing data in the provided data set, and if so, add an evolvable [Imputer](http://scikit-learn.org/stable/modules/generated/sklearn.preprocessing.Imputer.html) to the primitive set.

Parameters:

strategy: ["mean", "median", "most_frequent"]

rhiever:
@tonyfast, are you interested in taking on this issue before the 0.7 release in 2 weeks?

rhiever:
@teaearlgraycold, how's this coming along?

PR #269, opened 2016-09-10
Tonyfast/imputer

tonyfast:
## What does this PR do?

Add Imputer to the tpot preprocessors.

[`imputer.py`](https://github.com/tonyfast/tpot/blob/tonyfast/imputer/tpot/operators/preprocessors/imputer.py)

_No tests were added mostly because there weren't any for the `preprocessors`
#253
## Questions:
- Do the docs need to be updated?  _Yes._
- Does this PR add new (Python) dependencies? _No._

### DIFF
--- a/tpot/operators/preprocessors/__init__.py
+++ b/tpot/operators/preprocessors/__init__.py
@@ -32,3 +32,4 @@ from .polynomial_features import *
 from .robust_scaler import *
 from .standard_scaler import *
 from .zero_count import *
+from .imputer import *
--- a/tpot/operators/preprocessors/imputer.py
+++ b/tpot/operators/preprocessors/imputer.py
@@ -23,21 +23,29 @@ from sklearn.preprocessing import Imputer
 
 
 class TPOTImputer(Preprocessor):
-    """Uses scikit-learn's Imputer to transform the feature set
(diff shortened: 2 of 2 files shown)


teaearlgraycold:
A few things:

1. You'll need to add the Imputer to the preprocessor's `__init__.py` file so it is discovered by the operators module.
2. You should PR this onto the development branch.
3. You should add logic that makes it so the Imputer is always placed at the beginning of the pipeline if there are missing values in the input matrix. It probably doesn't ever need to be anywhere besides at the beginning of a pipeline, right?

rhiever:
👍 to @teaearlgraycold's suggestions.

tonyfast:
@rhiever @teaearlgraycold would the Imputer class handle the logic or would it go somewhere else?

tonyfast:
Should the Imputer even be an option in the pipeline?  It seems like it would only be required when the fitting step beings if `np.isnan(features).any()`.

rhiever:
I think the Imputer should only be added to the primitive set if there are any missing values in the data set.

rhiever:
Sorry it's taken me forever to get to this. We will also need a unit test verifying that a pipeline with this imputer work -- both data sets with and without missing data (two separate tests).

Issue #292, opened 2016-10-11
Add 'sag' solver to LogisticRegression operator for large dataset 


weixuanfu2016:
Add 'sag' solver to LogisticRegression operator for large dataset to speed up pipeline evaluation.
## Context of the issue

Default 'liblinear' solver works very good for normal dataset. But for the large dataset in the issue #271  and #290, it is very slow to evaluate pipelines with [LogisticRegression](http://scikit-learn.org/stable/modules/generated/sklearn.linear_model.LogisticRegression.html) operator. So the 'sag' solver (scikit-learn version > 0.17) should works better for large dataset. It is noted that `dual` argument should be `False` when using 'sag' solver.

Test code:

```
import time
import numpy as np
import matplotlib.pyplot as plt
from sklearn import datasets
from sklearn.model_selection import train_test_split
from sklearn.linear_model import LogisticRegression

heldout = [0.95, 0.90, 0.75, 0.50, 0.01]
rounds = 10

X, y = datasets.make_classification(n_samples=10000, n_features=4000,
                                    n_informative=2, n_redundant=10,
                                    random_state=42)

classifiers = [
    ("liblinear", LogisticRegression(solver='liblinear', tol=1e-1, C=1.e4 / X.shape[0])),
    ("sag", LogisticRegression(solver='sag', tol=1e-1, C=1.e4 / X.shape[0]))
]



xx = (1. - np.array(heldout) )*10000

for name, clf in classifiers:
    print("training %s" % name)
    rng = np.random.RandomState(42)
    yy = []
    for i in heldout:
        yy_ = []
        for r in range(rounds):
            X_train, X_test, y_train, y_test = \
                train_test_split(X, y, test_size=i, random_state=rng)
            time_start = time.time()
            clf.fit(X_train, y_train)
            time_end = time.time()
            yy_.append(time_end-time_start)
        yy.append(np.mean(yy_))
    plt.plot(xx, yy, label=name)

plt.legend(loc="upper right")
plt.xlabel("Training Size (N)")
plt.ylabel("Average Time Usage (s)")
plt.show()
```

![time_usage_liblinearvssag](https://cloud.githubusercontent.com/assets/21084970/19292133/e0cd7568-8fe7-11e6-9641-50af54344897.png)

rhiever:
Should we just always use the sag solver then?

weixuanfu2016:
[...]

Issue #271, opened 2016-09-13
"Fit" function sometimes never terminates 


choosehappy:
About 80% of the time tpot never finishes (>30 hours run time currently), and in the other 20% of the time, it finishes with the same code + parameters + input in ~5 hours and produces excellent results.

The code itself isn't anything exceptional:

import time
import numpy as np
from tpot import TPOT
train_data = np.genfromtxt('fold_1_trainFeatVec.csv', delimiter=',')
train_labels= np.genfromtxt('fold_1_trainLabels.csv', delimiter=',')

tpot = TPOT(generations=20, population_size=20)

start= time.time()
tpot.fit(train_data, train_labels)
print "train acc:\t %f" % tpot.score(train_data, train_labels)
end=time.time()

I've tried installing both via:
pip install tpot

as well as using this repo:
git clone [url]
python setup.py sdist
pip install --user TPOT-0.4.1.tar.gz

rhiever:
If you set `verbosity=2` in the `TPOT()` instantiation call, does it look like it's getting stuck on one pipeline evaluation, or is it just taking a long time in general?

Also, I strongly recommend upgrading to the latest version of TPOT (0.6.4). That alone may fix your issue.

choosehappy:
in the instances where it works, the progress bar does move at a fairly
consistent pace.

in the failing ones, it will randomly get stuck at some percent (usually <
50%) and then never move.

i was a bit curious about the version numbers, i'm glad you mentioned it.

this github master branch doesn't contain the latest code?

because when i used "pip install tpot" and it pulled it from a distant
repo, it did in fact load a 0.6.x version, but installing directly from
github results in 0.4.x?

rhiever:
The master branch has the latest stable release of TPOT, which is the same version that will install via pip. Hopefully the upgrade will fix your issues -- we made some major changes in 0.5 that should hopefully prevent TPOT from getting "stuck."

choosehappy:
Unfortunately, same issue after upgrading to the latest version 

it takes about 3 minutes to get to 8/420, and then remains there for 45 minutes afterwards

![image](https://cloud.githubusercontent.com/assets/9681868/18489581/5fcc321e-79fe-11e6-8b91-f27d54d5574c.png)


i've made the code and dataset available here:

http://gleason.case.edu/webdata/tpot/

rhiever:
[...]

PR #294, opened 2016-10-13
Add sag solver for logistic regression

weixuanfu2016:
## What does this PR do?

Add sag solver for logistic regression to handle large dataset.

tpot/operators/classifiers/logistic_regression.py

Any TPOT examples
#271
#292
#252
## Questions:

‘sag’ only handle L2 penalty and only works with dual = False. It means dual is False in both l1 and l2 penalty. Thus, no dual argument if sag is added. Is it OK?


weixuanfu2016:
This PR maybe not compatible with #267 since arguments' type format changed in that PR

Issue #303, opened 2016-11-04
Tpot tries the same Pipeline multiple times in population


ghgr:
I'm running some tests with tpot (_development_ branch) with some additional _print_ in order to view which pipeline is evaluated at each step. I found out that it often tries the same Pipeline (xgboost). Here is my output:

> Optimization Progress:   1%|          | 301/40040 [2:19:51<283:40:31, 25.70s/pipeline]Pipeline(steps=[('xgbregressor', XGBRegressor(base_score=0.5, colsample_bylevel=1, colsample_bytree=1, gamma=0,
>        learning_rate=0.14, max_delta_step=0, max_depth=7,
>        min_child_weight=20, missing=None, n_estimators=100, nthread=-1,
>        objective='reg:linear', reg_alpha=0, reg_lambda=1,
>        scale_pos_weight=1, seed=0, silent=True, subsample=1.0))])
> Score: 1228.471468
> Optimization Progress:   1%|          | 302/40040 [2:20:16<281:32:19, 25.51s/pipeline]Pipeline(steps=[('xgbregressor', XGBRegressor(base_score=0.5, colsample_bylevel=1, colsample_bytree=1, gamma=0,
>        learning_rate=0.14, max_delta_step=0, max_depth=7,
>        min_child_weight=20, missing=None, n_estimators=100, nthread=-1,
>        objective='reg:linear', reg_alpha=0, reg_lambda=1,
>        scale_pos_weight=1, seed=0, silent=True, subsample=1.0))])
> Score: 1228.471468
> Optimization Progress:   1%|          | 303/40040 [2:20:41<279:33:12, 25.33s/pipeline]Pipeline(steps=[('xgbregressor', XGBRegressor(base_score=0.5, colsample_bylevel=1, colsample_bytree=1, gamma=0,
>        learning_rate=0.14, max_delta_step=0, max_depth=7,
>        min_child_weight=20, missing=None, n_estimators=100, nthread=-1,
>        objective='reg:linear', reg_alpha=0, reg_lambda=1,
>        scale_pos_weight=1, seed=0, silent=True, subsample=1.0))])
> Score: 1228.471468
> Optimization Progress:   1%|          | 304/40040 [2:21:05<277:34:51, 25.15s/pipeline]Pipeline(steps=[('xgbregressor', XGBRegressor(base_score=0.5, colsample_bylevel=1, colsample_bytree=1, gamma=0,
>        learning_rate=0.14, max_delta_step=0, max_depth=7,
[...]

PR #316, opened 2016-11-16
Checking duplicated pipeline in TPOT optimization process

weixuanfu2016:
## What does this PR do?

In the source codes of [algorithms.eaSimplie](https://github.com/DEAP/deap/blob/master/deap/algorithms.py) in deap, the fitness.values were removed during generating offsprings in each generation by crossover or mutation. So these pipeline generated by crossover and mutation need be evaluated again in the new generation. However, if the diversity of pipelines are limited by customizing used operators. Many duplicated pipelines can be generated over generations. Evaluating the same pipelines is worthless. So in this RP, a list of evaluated individual is added. In each time of pipeline evaluation, the pipeline must be checked first whether it is evaluated before. If so, its fitness scores are assigned with that in previous evaluations. The way can speed up a lot TPOT pipeline optimization process in later generations. 

Line ~560 in base.py 

```

digits = load_digits()
                                                    train_size=0.10, test_size=0.90)

tpot = TPOTClassifier(generations=5, population_size=20, verbosity=3)
tpot.fit(X_train, y_train)

```

#303

### DIFF
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -561,15 +561,17 @@ class TPOTBase(BaseEstimator):
             # Fix random state when the operator allows
             self._set_param_recursive(sklearn_pipeline.steps, 'random_state', 42)
 
-            # Count the number of pipeline operators as a measure of pipeline covamplexity
+            # Count the number of pipeline operators as a measure of pipeline complexity
             operator_count = 0
 
-            #print('individual', individual)
-            #print('individual dir', dir(individual))
-            #print('individual fitness', individual.fitness.values)
+            # check if the individual are evaluated before
             if self.eval_ind.count(individual):
+                # get fitness score from previous evaluation
                 ind = self.eval_ind[self.eval_ind.index(individual)]
                 operator_count, resulting_score = ind.fitness.values
+                if self.verbosity == 3:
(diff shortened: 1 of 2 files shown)


weixuanfu2016:
Update: the list is changed to a dictionary which only store pipeline string and fitness stores to save the memory usage. I still keep this dictionary since not all the duplicate pipelines can be detected by parent-child comparison below
The `eaSimple` was replaced by adding built-in check duplicated pipeline by parent-child comparison.

weixuanfu2016:
One unit test in test.py only test one pipeline in TPOT, if the pipeline has a error then the unit test will fail. It is why the tests by Travis CI sometimes fail in a build but sometimes pass.

[Example 1](https://travis-ci.org/rhiever/tpot/builds/176604884)
[Example 2](https://travis-ci.org/rhiever/tpot/builds/176470237)

```
def test_fit():
    """Assert that the TPOT fit function provides an optimized pipeline"""
    tpot_obj = TPOTClassifier(random_state=42, population_size=1, generations=1, verbosity=0)
    tpot_obj.fit(training_features, training_classes)

    assert isinstance(tpot_obj._optimized_pipeline, creator.Individual)
    assert tpot_obj._gp_generation == 0
    assert not (tpot_obj._start_datetime is None)
```

Issue #310, opened 2016-11-12
Getting TPOT to work with sample_weights


joseortiz3:
Many (all?) of the classifiers and regressors (and probably some of the other pipeline elements) used by TPOT have a `sample_weights` functionality, which gives the relative importance of different samples during fitting.

I would expect the `sample_weights` feature to be used like
`
TPOTRegressor.fit(x_train, y_train, sample_weights=sample_weights)`

Which would result in the `sample_weights` being passed down to any pipeline model that can use it. For instance, if TPOT tries `GradientBoostingClassifier.fit()` (`GBC.fit()`), one of the arguments given to `GBC.fit()` would be the provided `sample_weights` (since GBC supports that already), so that the loss function used by `GBC` is weighted per-sample (not per class!). Any model that _could_ use `sample_weights` is now passed it.

How hard would it be to get this feature working? I tried to answer this question on my own, but it is challenging for someone new to the code.

**Motivation:**
I think sample weighting is important enough that it has become a standard in sklearn, so perhaps it should  also be incorporated into TPOT.

**Specific Use Case:**
In my work, my samples have very different relative importances, so this feature is crucial. My pipeline needs to predict a trinary classification: 0,1,2, for each sample (tens of thousands). But correctness is more important for some samples than others. For instance, it might be 10-1000x more important to correctly classify some samples than other samples.

weixuanfu2016:
Thank you for sharing this idea. I have some thoughts about it. Actually, it is related to a [old issue](https://github.com/scikit-learn/scikit-learn/issues/4632) in scikit-learn. The thing you may want to try is to wrap the cross_val_score from scikit-learn to including the sample weight. Here is a [demo](https://github.com/ndawe/scikit-learn/commit/3da7fb708e67dd27d7ef26b40d29447b7dc565d7).  It's a little hacky and not guaranteed to still work with the lastest version of scikit-learn.

rhiever:
@weixuanfu2016 / @sahilshah1194, can you please verify that all of the classifiers and regressors that we use in TPOT indeed take `sample_weights` into their `fit` functions? If so, it would be a straightforward API change for TPOT to allow `sample_weights` as an optional parameter to the `fit` function, which would simply pass those weights (or `None` by default) to the sklearn pipeline `fit` function.

weixuanfu2016:
[...]

PR #317, opened 2016-11-17
Adding sample_weight argument into tpot.fit()

weixuanfu2016:
## What does this PR do?

Make TPOT work with sample weights.

Line 527 in base.py

```

digits = load_digits()
                                                            train_size=0.10, test_size=0.90)

sample_weight = range(1, len(y_train) + 1) # made-up weights

tpot = TPOTClassifier(generations=5, crossover_rate= 0.5, population_size=20)
tpot.fit(X_train, y_train, sample_weight=sample_weight)
```

#310 

## Questions:

I wil add a unit test for this function soon.

### DIFF
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -495,7 +495,7 @@ class TPOTBase(BaseEstimator):
 
         return eval(sklearn_pipeline, self.operators_context)
 
-    def _set_param_recursive(self, pipeline_steps, parameter, value):
+    def _set_param_recursive(self, pipeline_steps, parameter, value, sample_weight = None):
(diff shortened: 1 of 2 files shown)


tpot/base.py
```diff
@@ -519,7 +523,31 @@ def _set_param_recursive(self, pipeline_steps, parameter, value):
                 if hasattr(obj, parameter):
                     setattr(obj, parameter, value)
 
-    def _evaluate_individual(self, individual, features, classes):
+    def _set_sample_weight(self, pipeline_steps, sample_weight):
```
rhiever: We shouldn't need a new function for this; we can use the existing `_set_param_recursive` function.

Issue #318, opened 2016-11-17
Refine the TPOT mutation operator


rhiever:
Currently, TPOT chooses one of three mutation operators when mutating a pipeline:

1) Shrink -- remove operator(s) from the pipeline
2) Point -- replace an existing operator with a randomly-generated operator OR change one of the operator's parameters
3) Grow -- replace an existing operator with 2+ randomly-generated operators

These mutations work well for TPOT's purposes, except for one issue: The point mutation can only replace the existing operator with a new operator that has the same number of parameters. i.e., if `Op1` has 2 integer parameters, it can't be replaced by `Op2` with 3 integer parameters nor `Op3` that has 1 float parameter. Not good!

We should verify the above claim, and if it's true, refine TPOT's mutation operators such that it's possible to do a true point mutation that can replace any operator with any other operator without changing the rest of the pipeline downstream.

PR #319, opened 2016-11-28
New point mutation operator

weixuanfu2016:
## What does this PR do?

A new point mutation operator is added and it replaces a randomly chosen primitive/terminal (operator/parameter). It is noted that the old point mutation operator mentioned in the issue #318 was `gp.mutUniform`, which is not exactly point mutation operator since it changes all subsequent steps in a pipeline  from the mutated points. The new point mutation operator can keep all the subsequent and only change one parameter or one operator (with randomly generating whole set of its parameters).

line 617 in base.py

Any TPOT examples.

#318 

## Questions:

As mentioned above, `gp.mutUniform` is not real point mutation function. But in the PR, I still keep in TPOT as one of 4 random mutation operators. I am not sure whether it should be removed.

Another concern is that the new point mutation can change the whole "pipeline" if the pipeline only has one operator and the mutated point is the primitive. The default mutation rate in TPOT is 0.9, which maybe too high to save good pipeline when generating a new generation by mutation.


rhiever:
Can you please bring this PR up to date with the latest of the `development` branch?

Also, maybe we should move all of the GP-related functions to `deap_gp.py`.

weixuanfu2016:
OK, make sense.

rhiever:
Latest commits seem to break. Something about a `NoneType` not being iterable.

weixuanfu2016:
Yep, I am looking into it. The reason is somehow the `self._optimized_pipeline` is NoneType. And both unit tests used the fit() with population_size = 1 and generations = 1. My guess is that a random pipeline failed and then was assign a `-float(inf)` score  but this error was not caught. I will fix the bug soon.

rhiever:
Yep, I suspect that is the error as well.

weixuanfu2016:
The bugs should be fixed.

PR #330, opened 2016-12-19
Fixed small problem when random_state = 0

Yatoom:
Small fix for checking if random_state is set, and documentation fix.

### DIFF
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -347,7 +347,7 @@ class TPOTBase(BaseEstimator):
 
             # Store the pipeline with the highest internal testing score
             if self._hof:
-                top_score = -5000.
+                top_score = -float('inf')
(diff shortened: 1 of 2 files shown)


rhiever:
Much appreciated! Can you please PR this to the `development` branch instead?

tpot/base.py
```diff
@@ -286,7 +286,7 @@ def fit(self, features, classes):
         features = features.astype(np.float64)
 
         # Set the seed for the GP run
-        if self.random_state:
+        if self.random_state is not None:
```
rhiever: Maybe this should be:

```python
if not (self.random_state is None):
```

Yatoom: I went to look up what the difference was. According to this thread hey compile to the same bytecode: http://stackoverflow.com/questions/2710940/python-if-x-is-not-none-or-if-not-x-is-none

rhiever: Works for me! 👍

Issue #327, opened 2016-12-15
Any way to speed up tpot ?


eromoe:
It is realy too slow for my work... 
I setup a 6 core 16G ram ubuntu vm, but I found tpot only use one core.

Does it (I don't see relative docs):

1. Support multi-core? 
2. Support multiple mechine cooperate? (may be something like [spark-sklearn](https://github.com/databricks/spark-sklearn) , I would like to help )
3. Support GPU?

Or is there any other way to speed up currentlly?

rhiever:
We've been looking at parallelization options but haven't found one that quite works for us yet. Perhaps the best option right now is to run multiple instances of TPOT (10+) to get a diversity of recommendations.

eromoe:
But multiple instances of TPOT may build many same pipeline,  why this way is better.

arita37:
@eromoe :  Maybe, you can try Ipypa ?

eromoe:
@arita37 What do you mean, any example ? Simply run multi tpot-instance would waste much calculation resource .The most easy and effcient way I can imagine is : 
Split all combinations of model and params tpot generated into several groups and run parallel. (But tpot not support now) .

arita37:
You can check ipyparall documentation, pretty straightforward

On 21 Dec 2016, at 16:48, eromoe <notifications@github.com> wrote:

@arita37 What do you mean, any example ? Simply run multi tpot-instance would waste much calculation resource .The most easy and effcient way I can imagine is :
Split all combinations of model and params tpot generated into several groups and run parallel. (But tpot not support now) .

―

eromoe:
I read the docs of ipyparall , it is just a distribute framework as my expect. If you want to do a big task efficiently, you must split it into pieces, then send to workers(ipyparall engine) which would excute tasks, but you must schedule tasks in right way first.  Runing tpot in multi-process is simple, but it not so effcient in that way.

brianmingus:
@rhiever A good way to do this would be to focus on algorithms that have parallel implementations, such as xgboost and sklearn SGD, setting n_jobs=-1. Looking at your GECCO '16 paper, I can make the following suggestions:

- Decomposition: Instead of RandomizedPCA, use sklearn.decomposition.SparsePCA and sklearn.linear_model.SGDClassifier. The latter may seem somewhat strange, but by creating an autoencoder with SGD you implement PCA.

[...]

Issue #302, opened 2016-11-04
Parallelize cross validation as a provisional optimization


ghgr:
I propose setting the parameter _n_jobs_ to _num_cv_folds_ to get a sort of quick parallelism. When better solutions with dask are implemented we could set it again to 1

On _base.py_, __evaluate_individual_ method (line 575)
`cv_scores = cross_val_score(self, sklearn_pipeline, features, classes, cv=self.num_cv_folds, scoring=self.scoring_function)`

to 

`cv_scores = cross_val_score(self, sklearn_pipeline, features, classes, cv=self.num_cv_folds, scoring=self.scoring_function, n_jobs=self.num_cv_folds)`

weixuanfu2016:
Thank you for sharing this nice tips. Based on the [User Guide](http://scikit-learn.org/0.17/modules/generated/sklearn.cross_validation.cross_val_score.html) of cross_val_score from  scikit-learn `njobs` parameter determines the CPU number to use in cross-validation while `cv` parameter determines the number of folds. Maybe adding `njobs` as another parameter in TPOT for paralleling the cross-validation with default of 1 since this way may use much more system resource. @rhiever

ghgr:
Indeed that would be more precise. I proposed to make _njobs == num_cv_folds_ since the default number of cv folds in **tpot** is 3, and most machines (used for machine learning) have more than 3 cores. Just to make @minimumnz feel better not having idle cores [1] ;-)

[1] https://github.com/rhiever/tpot/issues/177

rhiever:
We've been talking about adding a `n_jobs` parameter to TPOT for quite some time, which would basically do this. Perhaps we should just do that.

Jonathan-Livingston-Seagull:
Would not be better to use the parallel processing capabilities of DEAP. i.e Each combination(preprocessor, algorithm, postprocessor, etc) is an individual in the population. Hence exploiting the DEAP features helps TPOT parallelizing through running an individual in seperate cores?

rhiever:
We looked into using the multiprocessing capabilities of DEAP, but ran into issues with pickling lambda functions and a few other tricks we use in TPOT. Maybe @weixuanfu2016 can provide full details.

In the meantime, I've merged the PR into the `development` branch that exposes `n_jobs` for the cross-validation procedure.

PR #338, opened 2017-01-04
Paralleling pipeline evaluation

weixuanfu2016:
## What does this PR do?

The PR is for Paralleling pipeline evaluation in Linux and Mac. Instead of using n_jobs in `cross_cv_score`, this PR simply parallelizes the pipeline evaluation with a time limit in each evaluation, which can speed up a lot for TPOT.fit()  

```
import time

digits = load_digits()
                                                    train_size=0.25, test_size=0.75)

tpot = TPOTClassifier(generations=2, population_size=10, verbosity=2, max_eval_time_mins=0.02, n_jobs = 3, random_state = 42)
time_start = time.time()
tpot.fit(X_train, y_train)
print('\nTime used with num_cpu = 3:',time.time()-time_start)

```

#327 
#302 
#307 

## Questions:

Yes, it need `pathos` module


derekjanni:
Try adding `pip install pathos` [here](https://github.com/rhiever/tpot/blob/master/ci/.travis_install.sh#L54). Travis is complaining about the import, presumably because your lib's not being installed.

Also [this](https://github.com/weixuanfu2016/tpot/blob/4a6211fe35163aeff15217f53bc0d91a4d5ae0a1/tpot_test_multi_process.py 
) probably shouldn't exist at the top directory level. It should probably be added to `.gitignore`

weixuanfu2016:
Oh, thanks for your suggestions.

derekjanni:
Oh `pathos` needs to be [here](https://github.com/rhiever/tpot/blob/master/setup.py#L37) too.

Issue #296, opened 2016-10-16
New Feature : add Prelist of pipelines


arita37:
When we have worked on a class of datasetb (larget size, 1Go), we have an idea
of the optimal pipeline, and we just want to investigate the optimal parameters without computing the whole in genetic programm. It would reduce substantially the computation time.

For example:
  We can provide a list of pipeline with algorithm names ['LinearSVC', 'RF' ]
  We can provide a list of hyper-parameter for each algorithm to search for...

Main goal is to limit the search to those parameters.

rhiever:
This can currently be "hacked" into TPOT by modifying the [population instantiation procedure](https://github.com/rhiever/tpot/blob/master/tpot/base.py#L296), but it could be a nice feature to be able to do that through the regular TPOT interface. We'd have to put a lot of thought into how that would look. I'll file this as an enhancement request for now.

rhiever:
@teaearlgraycold is taking on this issue. He will be extending the `config_dict` parameter to support custom pipeline seeds.

hanshupe:
> This can currently be "hacked" into TPOT by modifying the [population instantiation procedure](https://github.com/rhiever/tpot/blob/master/tpot/base.py#L296), but it could be a nice feature to be able to do that through the regular TPOT interface. We'd have to put a lot of thought into how that would look. I'll file this as an enhancement request for now.

I would like to use the "hack", but I think the link is outdated. Could someone please update the Line number, where I can add the initial solutions? Thx!

weixuanfu:
I think it was refereed to [this line](https://github.com/EpistasisLab/tpot/blob/v0.11.5/tpot/base.py#L670)

hanshupe:
> I think it was refereed to [this line](https://github.com/EpistasisLab/tpot/blob/v0.11.5/tpot/base.py#L670)

Can you give me a hint how I can convert a predefined sklearn pipeline into a list of DEAP primitives that is needed for the population?

weixuanfu:
@hanshupe we don't have a example for this conversion . But we have [a example for reverse conversion](https://github.com/EpistasisLab/tpot/blob/master/tests/tpot_tests.py#L358-L370).

hanshupe:
Yes thanks I saw the compile method, but the conversion into the opposite direction is not so easy as I thought :)

PR #343, opened 2017-01-30
Refine operators and their parameters in TPOT

weixuanfu2016:
## What does this PR do?

1. removes the whole operators folder and replace it by using two python dictionaries (config_classifier.py and config_regessor.py) and a operator utils (operator_utils.py)

2. provides a new option `operator_dict_file`  (`--operator_dict` for command line) to easily customize list of operators and their parameters in TPOT using a operator dictionary file formatted as config_classifier.py. 

3. Adds a `_pre_test` decorator to decorate pipeline generator, crossover operator and mutation operator to make sure the new pipeline is workable with a small simulated dataset

4. fix > 15 unit tests for the new dynamically created operator classes. 

The operator utils (operator_utils.py)

Any tpot examples

#252 
#296 

## Questions:

May need refine selector and preprocessor's parameters in the dictionaries. 

`creator.Individual.from_string` in unit tests does not work with operators with parameter setting in dictionary because in this branch the parameter inside a operator is a independent parameter class but it may share same value. Then the function cannot tell this value belong to which class. If maybe a bug in deap module. It does not effect GP processes in TPOT since this function is not used in TPOT base.


weixuanfu2016:
Need debug for supporting python2.7

rhiever:
Cool -- let's go over this in-person tomorrow.

PR #403, opened 2017-03-30
Avoid using multiprocessing when n_jobs ==1

weixuanfu2016:
Fix 3 unit tests due to new dictionary 

multiprocessing can not be used when n_jobs==1

### DIFF
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -727,36 +727,30 @@ class TPOTBase(BaseEstimator):
                 resulting_score = -float('inf')
             return resulting_score
 
-        if self.n_jobs != 1:
-            pool = Pool(processes=self.n_jobs)
-            resulting_score_list = []
-            # chunk size for pbar update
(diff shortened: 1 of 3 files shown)


tpot/base.py
```diff
@@ -727,7 +727,7 @@ def _wrapped_cross_val_score(sklearn_pipeline, features=features, classes=classe
                 resulting_score = -float('inf')
             return resulting_score
 
-        if not sys.platform.startswith('win'):
+        if self.n_jobs != 1:
```
rhiever: It seems that there's a lot of repeated code between the two branches of this `if` statement. Is it possible to instead make the `if` statement decide what job mapper to use (1), then the rest of the code can be the same?

(1) It will use the job pool if n_jobs != 1, otherwise it will use regular `map`?

weixuanfu2016: check the new commit in the PR

weixuanfu2016:
I think this PR is not needed any more due to the PR #406. So I close it.

PR #415, opened 2017-04-13
Adding skrebate and mdr to setup and config_classification

sohnam:


rhiever:
A few requests:

1) Please send the PR to the development branch instead of the master branch.

2) Please make the TPOT-MDR config a separate config file

3) Re-work the dependencies to refer only to the package names: skrebate, scikit-mdr. You don't need to specify the classes within the packages.

PR #416, opened 2017-04-14
add requirements.txt file

iddoberger:


rhiever:
The requirements are processed in [setup.py](https://github.com/rhiever/tpot/blob/master/setup.py#L37); what's the advantage of having a separate `requirements.txt`?

iddoberger:
This requirements file is useful in order to preserve a working
environment, including dependencies.
In addition, installing the environment with  setup.py
<https://github.com/rhiever/tpot/blob/master/setup.py#L37> also install the
current project as a dependency.
Although it seem like a duplication, it is common practice to have both. As
seen here:
https://github.com/explosion/spaCy/blob/master/requirements.txt
https://github.com/explosion/spaCy/blob/master/setup.py#240

rhiever:
Interesting. I have a couple more questions:

1) Is it possible to specify optional dependencies within the `requirements.txt`?

2) Could this PR also add code to `setup.py` that parses `requirements.txt` to build the dependency list, to avoid duplication of information?

iddoberger:
1. I have seen projects that use a file called "optional-requirements.txt"
for that purpose.

2. It is possible to add code that does that:
http://stackoverflow.com/a/14399775/72099
 But I am not sure if it is the right thing to do:
https://caremad.io/posts/2013/07/setup-vs-requirement/
http://www.cdotson.com/2015/08/dont-use-parse_requirements-in-your-code/

rhiever:
We'll look into this---thanks!

PR #417, opened 2017-04-14
Use ExtraTreesClassifier in RFE and add lite operator dictionary 

weixuanfu2016:
1. Replace SVC esitmator in RFE by ExtraTreesClassifier

2. Add Lite operator dictionary based on the issue #384 

3. Add parameter in TPOT for using lite mode (`dict_lite` in TPOTClassifier or TPOTRegessor and `--lite` in command line)

4. Add unit tests and documentation for new prarameter

PR #418, opened 2017-04-17
tpot_mdr_classifier and setup

sohnam:


setup.py
```diff
@@ -35,7 +35,9 @@ def calculate_version():
 ''',
     zip_safe=True,
     install_requires=['numpy', 'scipy', 'scikit-learn>=0.18.1', 'deap', 'update_checker', 'tqdm', 'pathos'],
-    extras_require={'xgboost': ['xgboost']},
+    extras_require={'xgboost': ['xgboost'],
+                   'skrebate': ['skrebate'],
+                   'mdr': ['scikit-mdr']},
```
rhiever: I think this line should be ```'scikit-mdr': ['scikit-mdr']},```

Can @weixuanfu2016 or @teaearlgraycold confirm?

tpot/tpot_mdr_classifier.py
```diff
@@ -0,0 +1,30 @@
+import numpy as np
```
rhiever: Please add the header that the other configuration dictionaries have.

tpot/tpot_mdr_classifier.py
```diff
@@ -0,0 +1,30 @@
...
+
+tpot_mdr_classifier_config_dict = {
+
+    # Classifiers
+    
+    'mdr.MDR': {
+    'tie_break': tie_break % 2
```
rhiever: These should be value entries, not preprocessing code. So this line would be, e.g.,

```'tie_break': [0,1],```

File: tpot/tpot_mdr_classifier.py [hunk 4]
```diff
@@ -0,0 +1,30 @@
...
+    'default_label': default_label % 2
+    },
+    
+    # Feature Selectors
+    
+    'skrebate.ReliefF': {
+        'n_features_to_select': max(1, n_features_to_select)
```
rhiever: Another example, this line would be:

```'n_features_to_select': range(1, 6)```

rhiever:
I sorted out the merge conflict in `setup.py`.

@weixuanfu2016, please keep this PR in mind for #417.

tpot/tpot_mdr_classifier.py
rhiever: The ReliefF algorithm should also include the `n_neighbors parameter`, with the values

```[2, 10, 50, 100, 250, 500]```

tpot/tpot_mdr_classifier.py
rhiever: Please also rename this file to `config_mdr_classifier.py`.

PR #428, opened 2017-04-27
Add Imputer

teaearlgraycold:
## What does this PR do?

* Imputes missing values (`nan`s) in feature sets if found.
* Adds tests for the imputer


tpot/base.py
```diff
@@ -392,27 +392,19 @@ def fit(self, features, classes, sample_weight=None):
...
 
-        # Check that the input data is formatted correctly for scikit-learn
-        if self.classification:
-            clf = DecisionTreeClassifier(max_depth=5)
-        else:
-            clf = DecisionTreeRegressor(max_depth=5)
+        if self._contains_nan(features):
+            if self.verbosity > 1:
+                print('Imputing missing values in feature set')
+            features = Imputer(strategy='most_frequent').fit_transform(features)
```
rhiever: This should use mode for discrete features and median for continuous features.

tpot/base.py
```diff
@@ -541,7 +533,15 @@ def predict(self, features):
...
         if not self._fitted_pipeline:
             raise RuntimeError('A pipeline has not yet been optimized. Please call fit() first.')
-        return self._fitted_pipeline.predict(features.astype(np.float64))
+
+        features = features.astype(np.float64)
+
+        if self._contains_nan(features):
+            if self.verbosity > 1:
+                print('Imputing missing values in feature set')
+            features = Imputer(strategy='most_frequent').fit_transform(features)
```
rhiever: Maybe these 4 lines should be a function into themselves, since they're repeated >1 times?

tpot/base.py
```diff
@@ -641,6 +641,47 @@ def export(self, output_file_name):
...
+            Feature matrix
+        classes: array-like {n_samples}
+            List of class labels for prediction
+
+        Returns
+        -------
+        None
+        """
+        try:
+            check_array(features, dtype=np.float64, accept_sparse='csc')
```
rhiever: `accept_sparse` should be `None` (IIRC) because we do not currently allow sparse matrices in TPOT.

tpot/base.py
```diff
@@ -641,6 +641,47 @@ def export(self, output_file_name):
...
+            check_array(classes, ensure_2d=False, dtype=None)
+
+            assert len(classes.shape) == 1
+            assert classes.shape[0] == features.shape[0]
+        except Exception:
```
rhiever: Is there a more specific exception we can catch here instead?

tpot/base.py
```diff
@@ -641,6 +641,47 @@ def export(self, output_file_name):
...
+
+        Parameters
+        ----------
+        features: array-like {n_samples, n_features}
+
+        Returns
+        -------
+        bool, True if NaN is present, False otherwise.
+        """
+        return np.isnan(np.sum(features))
```
rhiever: Are you *sure* that `np.isnan(np.sum(features))` is faster than `np.any(np.isnan(features))`?

tests.py
```diff
@@ -454,6 +454,39 @@ def test_fit2():
...
+        population_size=1,
+        offspring_size=2,
+        generations=1,
+        verbosity=0,
+        config_dict='TPOT light'
+    )
+    features_with_nan = np.copy(training_features)
+    features_with_nan[0][0] = float('nan')
+
+    tpot_obj.fit(features_with_nan, training_classes)
```
rhiever: No `assert`?

tests.py
```diff
@@ -454,6 +454,39 @@ def test_fit2():
...
+    tpot_obj.predict(features_with_nan)
```
rhiever: No `assert`?

docs/using/index.html
```diff
@@ -610,6 +612,9 @@ <h2 id="customizing-tpots-operators-and-parameters">Customizing TPOT's operators
 
 <p>For more detailed examples of how to customize TPOT's operator configuration, see the default configurations for <a href="https://github.com/rhiever/tpot/blob/master/tpot/config_classifier.py">classification</a> and <a href="https://github.com/rhiever/tpot/blob/master/tpot/config_regressor.py">regression</a> in TPOT's source code.</p>
 <p>Note that you must have all of the corresponding packages for the operators installed on your computer, otherwise TPOT will not be able to use them. For example, if XGBoost is not installed on your computer, then TPOT will simply not import nor use XGBoost in the pipelines it considers.</p>
+<p><a name="usagenotes"></a></p>
+<h2 id="usage-notes">Usage Notes</h2>
+<p>When feature matrices are passed into TPOT, any "NaN" values will automatically be imputed. The imputation process uses the most frequent value from each feature as its best guess for each missing value.</p>
```
rhiever: Please replace `"NaN"` with `missing`

rhiever:
Ping.

tests.py
teaearlgraycold: The test aim's to check if the function will raise a `ValueError`, that alone will cause the test to fail. I've updated the test description to better reflect this.

tpot/base.py
teaearlgraycold: Looks like the difference is negligible. `np.any(np.isnan())` is more self-expanatory, so I've replaced the `_contains_nan` function with that function chain.

tpot/base.py
teaearlgraycold: No, assert will throw an `AssertionError` and `check_array` will throw a `ValueError`.

teaearlgraycold:
I apologize, I'd left this open by mistake.

tpot/base.py
teaearlgraycold: My bad, just learned you can catch multiple exception types on one line like so:

```python
except (AssertionError, ValueError):
    # foobar
```

tests.py
weixuanfu2016: I think the structure `try...except` below is better and consistent with other expectation tests, like `test_predict_proba2` 
```
try:
    #something without error
    assert True
except ValueError:
    assert False
```

tests.py
weixuanfu2016: I think the structure `try...except` mentioned above is better and more specific to one line or few line that may raise a Exception.

tests.py
```diff
@@ -454,6 +454,56 @@ def test_fit2():
...
+
+
+def test_imputer3():
```
weixuanfu2016: See comments above for adding `assert` in unit tests

tpot/base.py
```diff
@@ -392,27 +392,17 @@ def fit(self, features, classes, sample_weight=None):
...
+            features = self._impute_values(features)
 
-        try:
-            clf = clf.fit(features, classes)
-        except Exception:
-            raise ValueError('Error: Input data is not in a valid format. '
-                             'Please confirm that the input data is scikit-learn compatible. '
-                             'For example, the features must be a 2-D array and target labels '
-                             'must be a 1-D array.')
+        self._check_dataset(features, classes)
```
weixuanfu2016: why not use [check_X_y](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/utils/validation.py#L435)? it can be imported by `from sklearn.utils import check_X_y`

Another benefits from `check_X_y` is that the return value of y are converted to `np.float64` by [this line](https://github.com/scikit-learn/scikit-learn/blob/master/sklearn/utils/validation.py#L532). So `features = features.astype(np.float64)` in self.fit() could be deleted then.

tests.py
teaearlgraycold: In my merged code cleanup PR I removed those assertions as I considered them redundant.

`assert True` has no effect on the behavior of the test, and

```python
except ValueError:
    assert False
```

is just trading a `ValueError` for an `AssertionError`. Since no additional functionality is provided, and the `try ... except` pattern is just used to hint at what the test is checking, it should suffice to stick a comment next to the function call. I already include that information in the test function's docstring, but it could be repeated.

tpot/base.py
teaearlgraycold: You're right - that is a better option.

tests.py
weixuanfu2016: Hmm, make sense.

rhiever:
Looks like this PR has merge conflicts now too.

tpot/base.py
```diff
@@ -684,6 +678,61 @@ def export(self, output_file_name):
...
+            feature: array-like {n_samples}
+                An individual feature from a feature matrix.
+
+            Returns
+            -------
+            array-like {n_samples}
+            """
+            return Imputer(strategy="median", axis=1).fit_transform(feature)[0]
+
+        return np.apply_along_axis(impute_feature, axis=0, arr=features)
```
rhiever: This part isn't necessary. `Imputer` is already applied to each column separately.

teaearlgraycold: Fixed.

docs_sources/using.md
```diff
@@ -517,3 +517,8 @@ When using the command-line interface, the configuration file specified in the `
 For more detailed examples of how to customize TPOT's operator configuration, see the default configurations for [classification](https://github.com/rhiever/tpot/blob/master/tpot/config_classifier.py) and [regression](https://github.com/rhiever/tpot/blob/master/tpot/config_regressor.py) in TPOT's source code.
 
 Note that you must have all of the corresponding packages for the operators installed on your computer, otherwise TPOT will not be able to use them. For example, if XGBoost is not installed on your computer, then TPOT will simply not import nor use XGBoost in the pipelines it considers.
+
+<a name="usagenotes"></a>
+## Usage Notes
```
rhiever: Maybe change this to "Missing values"

and add a note that if the user doesn't want to use median imputation, then they should apply their own imputation method prior to passing the data to TPOT.

tpot/base.py
```diff
@@ -584,7 +572,13 @@ def predict(self, features):
         """
         if not self._fitted_pipeline:
             raise RuntimeError('A pipeline has not yet been optimized. Please call fit() first.')
-        return self._fitted_pipeline.predict(features.astype(np.float64))
+
+        features = features.astype(np.float64)
+
+        if np.any(np.isnan(features)):
+            features = self._impute_values(features)
```
rhiever: When predicting, TPOT should use the same Imputer that it trained on the training data. So TPOT needs to internally store the trained Imputer from the `fit` call.

rhiever:
@teaearlgraycold, think you can push the requested changes before this upcoming Monday?

PR #439, opened 2017-05-02
Treat custom config files as modules

teaearlgraycold:
## What does this PR do?

When a path is specified in the `config_dict` parameter of `TPOTBase` the file is read and parsed as a Python module, rather than parsed as a regular language and `eval`'d. This adds the restriction of requiring a consistent naming scheme for the configuration file. In this PR I've decided that as `tpot_config`.

In the `_read_config_file` function in `base.py`.

## Questions:

Yes, and I've updated them.


tpot/base.py
```diff
@@ -298,19 +299,35 @@ def _setup_config(self, config_dict):
...
 
+    def _read_config_file(self, config_path):
+        try:
+            custom_config = imp.new_module('custom_config')
+
+            with open(config_path, 'r') as config_file:
+                file_string = config_file.read()
+                exec(file_string, custom_config.__dict__)
+
+            self.config_dict = custom_config.tpot_config
```
weixuanfu2016: Very nice function. I think maybe returning the `custom_config.tpot_config` and assign `self.config_dict` in `_setup_config()` above is a little bit more consistent in structure.

teaearlgraycold: Good suggestion.

Issue #422, opened 2017-04-24
Too long idling of multiprocessing between each generation in TPOT 0.7.2 or 0.7.1


YukiSakai1209:
When I analyze my data using TPOT 0.7.2 or 0.7.1, it somehow works, but sometimes I have to wait too much idling time (stop of multiprocessing) between each generation. It often costs only about 5-10 minutes but sometimes up to several hours. I analyzed same data several times, it happened irregularly. For example, if I set max_time_mins = 480, sometimes 13000 pipelines have finished but sometimes only 2000 pipelines. It is too slow sometimes, but it seems not to stop since it restarts multiprocessing several hours later. During idling time only 1 thread is working.

Example: idling between 1:41:07 and 3:09:10, 3:22:39 and 6:33:23
Optimization Progress: 100%|| 200/200 [16:43<00:00, 117.68s/pipeline]Generation 1 - Current best internal CV score: 0.6093781690008105
Optimization Progress: 100%|| 300/300 [17:59<00:00, 58.66s/pipeline]Generation 2 - Current best internal CV score: 0.6093781690008105
Optimization Progress: 100%|| 400/400 [19:11<00:00, 29.34s/pipeline]  Generation 3 - Current best internal CV score: 0.6093781690008105
Optimization Progress: 100%|| 500/500 [22:27<00:00, 16.38s/pipeline]Generation 4 - Current best internal CV score: 0.6093781690008105
Optimization Progress: 100%|| 600/600 [25:13<00:00,  9.43s/pipeline]Generation 5 - Current best internal CV score: 0.6093781690008105
Optimization Progress: 100%|| 700/700 [30:41<00:00,  7.01s/pipeline]Generation 6 - Current best internal CV score: 0.6093781690008105
Optimization Progress: 100%|| 800/800 [38:33<00:00,  8.67s/pipeline]Generation 7 - Current best internal CV score: 0.6135697635697636
Optimization Progress: 100%|| 900/900 [41:21<00:00,  6.60s/pipeline]Generation 8 - Current best internal CV score: 0.6171742094383603
Optimization Progress: 100%|| 1000/1000 [47:07<00:00,  7.06s/pipeline]Generation 9 - Current best internal CV score: 0.6171742094383603
Optimization Progress: 100%|| 1100/1100 [51:27<00:00,  6.82s/pipeline]Generation 10 - Current best internal CV score: 0.6171742094383603
[...]

PR #440, opened 2017-05-02
Less freezing time in multiprocessing with large arrays

weixuanfu2016:
## What does this PR do?

1. Change the backend of the parameter `max_eval_time_mins` that controls how many minutes TPOT has to optimize a single pipeline. The backend is based on [stopit module](https://github.com/glenfant/stopit)

2. Change the backend of multiprocessing to dask instead of joblib

3. Fix a old issue in Windows. Now TPOT allow Control+C during optimization process when n_job !=1 in Windows

4. A new unit test for timeout function

5. Add `check_X_y` for checking dataset format. 

6. Update documentation

[base.py](https://github.com/weixuanfu2016/tpot/blob/joblib_timeout/tpot/base.py#L769-L793)

```
# coding: utf-8
from sklearn.datasets import make_classification
# make a huge dataset
X, y = make_classification(n_samples=50000, n_features=200,
                                    n_informative=20, n_redundant=20,
                                    n_classes=5, random_state=42)

# max_eval_time_mins=0.1 means 6 seconds limits for evaluating a single pipeline 
tpot = TPOTClassifier(generations=5, population_size=50, offspring_size=100, random_state=42, n_jobs=2, max_eval_time_mins=0.1, verbosity=3) 
tpot.fit(X, y)
```

[parallel processes freezing when matrices are too big](https://github.com/joblib/joblib/issues/138)
[Joblib hangs without crashing](https://github.com/joblib/joblib/issues/125)

#436 #422 

## Questions:

- Do the docs need to be updated?  Yes, docs are already updated in the PR.
- Does this PR add new (Python) dependencies? Yes, `stopit` and `dask`


rhiever:
I think the best workaround (for now) is to make TPOT *not* use multiprocessing when `n_jobs=1`, and put warnings in the documentation that enabling multiprocessing (`n_jobs!=1`) may be slow and prone to freezing with very large datasets.

sklearn has this same problem, right? e.g. if you use `cross_val_score` with `n_jobs!=1` for a very large dataset, it will also be slow and/or freeze.

weixuanfu2016:
Yep, sklean has a open issue for [this](https://github.com/scikit-learn/scikit-learn/issues/5115)

weixuanfu2016:
Although I set TPOT not use joblib when n_jobs=1, I found that many sklean's classes and functions used the built-in joblib and then freeze happened again.

tpot/base.py
```diff
@@ -771,24 +765,32 @@ def _evaluate_individuals(self, individuals, features, classes, sample_weight=No
...
+                        sklearn_pipeline,
+                        features,
+                        classes,
+                        self.cv,
+                        self.scoring_function,
+                        sample_weight,
+                        timeout=self.max_eval_time_seconds
+                    )
+                    jobs.append(job)
+                tmp_scores = compute(*jobs, get=multiprocessing.get, num_workers=self.n_jobs)
```
mrocklin: There might be some benefit to checking in with Dask's globals before hard-wiring `dask.multiprocessing.get`.  You might try the following instead:

```python
import dask
if njobs == 1:
    get = dask.async.get_sync
elif 'get' in dask.context._globals:
    get = dask.contest._globals['get']
else:
    get = multiprocessing.get

compute(..., get=get)
```

This would allow users to step in with other schedulers, like the distributed scheduler.  Also if you are only using a single core then it's best to avoid the multiprocessing scheduler, which has some non-trivial data movement overhead.

mrocklin:
Also, apologies for going off topic, but if you all ever wanted to dive more deeply into Dask you might consider the asynchronous concurrent.futures interface, notably [submit](http://distributed.readthedocs.io/en/latest/api.html#distributed.client.Client.submit) and [as_completed](http://distributed.readthedocs.io/en/latest/api.html#distributed.client.as_completed).  Here is an example of a somewhat-related library that uses this interface well to solve a somewhat-related computational problem: https://github.com/eriknw/dask-patternsearch .

Sorry for the intrusion.  I ran across this PR while searching github.

tpot/base.py
weixuanfu2016: Thanks for you suggestions.

weixuanfu2016:
@mrocklin No apology needed. I am very happy to find better solution. Thanks again for these good suggestions. Right now, the main issue is that the freezing with a large dataset seems from the scikit-learn and its joblib as mentioned here and #436 , which is very hard to solve until sklearn fixed it.  We used a thread-based timeout decorator for trying to kill these freezing jobs but somehow it is not works very well. The decorator can kill the freezing jobs but not under the right time limit. I suspect it is due to GIL. I will look into the example to check if it change solve our issues, Let me know if you have any idea about it.

rhiever:
Looks like this PR has merge conflicts now too.

weixuanfu2016:
OK, I will resolve these conflicts in `tests.py`

tpot/base.py
```diff
@@ -449,6 +444,16 @@ def fit(self, features, classes, sample_weight=None):
                     'a more reasonable outcome from optimization process in TPOT.'
                     )
 
+        if (features.shape[0] > 10000 or features.shape[1] > 100) and self.n_jobs !=1:
```
rhiever: Warnings should only print if verbosity>2.

weixuanfu2016: Make sense. I fixed it.

rhiever:
Looks like something broke with the rest of the merges. :-(

weixuanfu2016:
Conflicts fixed

weixuanfu2016:
I also add the patches in master branch (0.7.5) in this PR

Issue #388, opened 2017-03-23
Add subsample option for speeding up TPOT at large dataset


weixuanfu2016:
Adding a subsample option to set the subsample ratio of the training instance. For example setting it to 0.5 means that TPOT randomly collected half of the training data instances to evaluate a pipeline.

But first we need check the effects on accuracy from this option before adding it. This issue is related to #87

weixuanfu2016:
![training_sample_size_on_percentage_decrease_in_accuracy_based_on_pmlb](https://cloud.githubusercontent.com/assets/21084970/25672938/a34ef4b6-3003-11e7-8548-8307bc118899.png)

The subsample option were tested on >160 benchmarks in [PMLB](https://github.com/EpistasisLab/penn-ml-benchmarks/tree/master/pmlb). Subsample ratios were set from 0.1 to 0.9 with step = 0.1. The percentage changes in accuracy with partial training datasets were calculated based on accuracy with full dataset in each benchmark.  The scatter plot above shows the relationship between training sample size and percentage decreases in accuarcy (the results with training sample size > 5000 are excluded for clearer presentation in the scatter plot). It seems that once train sample size is less then ~1500, the changes in accuracy is very random and could be higher than 30%.  Based on this issue, I think I will add a friendly warning message for using `subsample` option once the training sample size is less then 1500 just in case that the option is used by mistake.

PR #441, opened 2017-05-03
Add subsample option in TPOT

weixuanfu2016:
## What does this PR do?

1. Add `subsample` option and it is subsample ratio of the training instance. For example, setting it to 0.5 means that TPOT randomly collects half of training samples for pipeline optimization process.

2. This function are designed for very large dataset. So a warning message is raised once the subsample size is too small. Too small subsample size cause unpredictable outcomes as mentioned in issue #388 

3. Add unit tests and docs for this function

base.py

Any TPOT example with adding `subsample` parameter.

#388 

#388 

## Questions:

- Do the docs need to be updated? Yes, updated in the PR


weixuanfu2016:
OK, I will resolve these conflicts in `tests.py`

rhiever:
Merged.

weixuanfu2016:
...Miss a commit, Merget too quickly/

rhiever:
Merged that one too.

weixuanfu2016:
Thanks!

Issue #450, opened 2017-05-16
config_dict='TPOT MDR' Error


ilyes495:
[provide general introduction to the issue and why it is relevant to this repository]
I'm trying to run TPOT classifier on my dataset using MDR configuration option, but I get the following error 
![screenshot from 2017-05-16 17-24-20](https://cloud.githubusercontent.com/assets/17175391/26111077/c8947238-3a5c-11e7-9674-715a28987d03.png)
 .........
![screenshot from 2017-05-16 17-27-29](https://cloud.githubusercontent.com/assets/17175391/26111137/f784096e-3a5c-11e7-822e-bea02b6e17f2.png)

Any help with the problem........ Thanks

weixuanfu2016:
It seems that no valid pipeline was randomly generated. Could you please let me know which version of scikit-MDR? I think `TPOT MDR` should work with version 0.4.3 and 0.4.4.

ilyes495:
scikit-MDR was 0.4.4  and I changed it to 0.4.3 and still give the same error

weixuanfu2016:
Thank you for letting us know this issue. The issue is caused by a recent update in [scikit-MDR](https://github.com/EpistasisLab/scikit-mdr/commit/6046ac963000de6ba90d2c0194a2a7446ea4b755#diff-102b1487e5185df93069d2af9ffce457) where MDR was not setting as a subclass of `ClassifierMixin` .

I will submit a PR soon to fix this compatibility issue in TPOT. 

For solving the issue temperately, please try the codes below to use TPOT MDR.

```
tpot_mdr_classifier_config_dict = {

    # Classifiers

    'mdr.MDRClassifier': {
        'tie_break': [0, 1],
        'default_label': [0, 1]
    },

    # Feature Selectors

    'skrebate.ReliefF': {
        'n_features_to_select': range(1, 6),
        'n_neighbors': [2, 10, 50, 100, 250, 500]
    },

    'skrebate.SURF': {
        'n_features_to_select': range(1, 6)
    },

    'skrebate.SURFstar': {
        'n_features_to_select': range(1, 6)
    },

    'skrebate.MultiSURF': {
        'n_features_to_select': range(1, 6)
    }

}

tpot = TPOTClassifier(generations=5, population_size=50, verbosity=3, config_dict=tpot_mdr_classifier_config_dict)

```

weixuanfu2016:
Or you can try this dictionary in dev branch.
```
[...]

PR #451, opened 2017-05-16
A hot patch for TPOT MDR for supporting latest version of scikit-MDR

weixuanfu2016:
1. Fix compatibility issue in TPOT MDR with the lastest version (0.4.4) of scikit-MDR

2. Add a unit test for TPOT MDR

Related issue #450

### DIFF
--- a/ci/.travis_install.sh
+++ b/ci/.travis_install.sh
@@ -53,6 +53,8 @@ fi
 
 pip install update_checker
 pip install tqdm
+pip install scikit-mdr
+pip install skrebate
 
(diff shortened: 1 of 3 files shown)


weixuanfu2016:
Note: this PR is just a patch and should be merged into master batch. This compatibility issue is fixed in dev batch. So for merging with dev batch later, the change in [7cea3bf](https://github.com/rhiever/tpot/pull/451/commits/7cea3bf8d1059fe15ef3e7fe50c748f65bfe0a2e) can be deleted but unit test in this PR is helpful.

PR #482, opened 2017-06-02
Update the driver for 0.8 release

rhiever:
`driver.py` wasn't updated for the release. As support for the CLI driver is low-priority, we can release this with any other fixes in the 0.8.1 release.

### DIFF
--- a/tpot/driver.py
+++ b/tpot/driver.py
@@ -228,10 +228,10 @@ def _get_arg_parser():
             'f1_micro, '
             'f1_samples, '
             'f1_weighted, '
-            'log_loss, '
-            'mean_absolute_error, '
-            'mean_squared_error, '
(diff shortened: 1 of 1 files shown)


weixuanfu2016:
Hmm, nice catch. Maybe it can be a patch for 0.8.0. I think the bug will cause issues when verbosity>=3 in command line mode.

rhiever:
True. But let's just wait to see if any other bug reports pop up in the next few days. For example, I've had TPOT freezing on me a lot when I run it from the CLI with `n_jobs!=1`, even with smaller datasets. Multiprocessing has me worried that it's causing too many freezing issues in TPOT.

weixuanfu2016:
Hmm, How about the PR #440? Maybe it can solve the issue.

Issue #484, opened 2017-06-02
Pipeline optimizer export uses old imports


KhaledSharif:
## Context of the issue

The ```pipeline_optimizer.export``` function uses an old import when it exports the pipeline to a Python file. Namely, the ```StackingEstimator``` class. (TPOT v0.8.0)

## Process to reproduce the issue

```pipeline_optimizer.export('tpot_exported_pipeline.py')```

## Expected result

```from tpot.builtins import StackingEstimator```

## Current result

```from tpot.built_in_operators import StackingEstimator```

## Possible fix

Correct the [following line](https://github.com/rhiever/tpot/blob/master/tpot/export_utils.py#L208) in the ```export_utils.py``` file to use the right import.

weixuanfu2016:
Hmm, you are right! It need to be updated since we moved these built-in operators to the `builtins ` folders. I will make a patch soon!

weixuanfu2016:
PR for the patch is posted! It maybe merged to master branch later. You may try my fixed branch just for now.

rhiever:
Whoops! Guess that one got overlooked in @teaearlgraycold's reorg PR. This fix will probably roll out on to master/pip later today.

PR #485, opened 2017-06-02
A patch for fixing a bug in issue 484

weixuanfu2016:
#484 

A quick patch for fixing a bug in export_utils.py mentioned in issue 484


rhiever:
Maybe it'd be a good idea (in a separate PR) to add a unit test importing all of those built-in import statements in the export module?

PR #488, opened 2017-06-02
Minor improvements

PG-TUe:
## What does this PR do?
Slightly refactored the TPOTBase._evaluate_individuals function changing:
 - Better legibility. After the introduction of parallelized evaluations of individuals, there was a lot of list-keeping added. This made the function very verbose and made it harder to see from a glance what happened. The rework requires zero list-indexing or sorting and has fewer lists, which hopefully makes it easier to read. Furthermore, some logic used to be delayed (eg. faulty pipelines would only be recorded in the evaluated_individuals_ dict way after they were found to be faulty, operator_count would first be calculated, but then quickly altered (max-operator) much later, just before being stored), these things are now more centralized.
- While it was checked if individuals had been evaluated before, if two identical new individuals would be asked to be evaluated, both would be (double work). Now, the list of all individuals is first reduced to a list of unique individuals before doing all the work, so no identical individual has to be evaluated twice.

All changes are confined to _evaluate_individuals.

Just run a before and after version, it should behave exactly the same (minus a slight performance improvement due to no duplicate evaluations).
One of the tests explicitly verifies that the ordering of input and output is the same (and it passes).
**note**: One test fails for me, (assert that ZeroCount operator returns correct transformed X), but this **also** happens in the development branch. So as far as I can tell, no errors have been introduced.

These are some changes I have been using in my own branches for a while now, I figured I would add them here instead of waiting for an eventual Gradual TPOT integration.

I did not submit a separate issue for this.

### DIFF
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -842,7 +842,6 @@ class TPOTBase(BaseEstimator):
         unique_individuals = [ind for i, ind in enumerate(individuals) if i in unique_individual_indices]
 
         # return fitness scores
-        fitnesses_dict = {}
         operator_counts = {}
         # 4 lists of DEAP individuals, their sklearn pipelines and their operator counts for parallel computing
         eval_individuals_str = []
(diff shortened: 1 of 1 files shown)


weixuanfu2016:
Looks very nice!

rhiever:
Looks great---thanks @PG-TUe!

Issue #490, opened 2017-06-02
v0.8.2 TPOTRegressor terminates prematurely


ap307:
Tried running the Boston housing price example on TPOT v0.8.2 in Python 3.6.

The fit() function runs for a few seconds, reaches a few percentage points, and then terminates. No pipeline is successfully fitted.

Code being run:

```
from tpot import TPOTRegressor
from sklearn.datasets import load_boston
from sklearn.model_selection import train_test_split

digits = load_boston()
X_train, X_test, y_train, y_test = train_test_split(digits.data, digits.target,
                                                    train_size=0.75, test_size=0.25)

tpot = TPOTRegressor(generations=5, population_size=50, verbosity=2)
tpot.fit(X_train, y_train)
```

rhiever:
Can you please provide more details on your Python environment and OS? That example is running fine on my machines.

ap307:
Running Python 3.6.1 on Windows 10. Appreciate the help.

ap307:
Got the Boston housing price example to work after uninstalling xgboost v0.6. Appears that it's the interaction with xgboost that is causing the issue? Had installed it using pip install xgboost.

ap307:
OK, I think the issue is resolved. Unfortunately v0.6a2 is not available on WIndows 10 as far as I can see at this stage.

rhiever:
TPOT has been a little finnicky with XGBoost, but the latest release should work with XGBoost 0.6a2. I just committed a patch that should add support in for 0.6. There's a weird issue where XGBoost stores both a `seed` and `random_state` parameter, both of which do the same thing, and XGBoost throws an error if they're not equal. Good times... thankfully they're fixing it soon.

BTW: You can install XGBoost on Anaconda now: `conda install py-xgboost`

ap307:
Great, this is working now (on a source install of xgboost v0.6 with GPU support).

Any plans to enable use of XGboost's GPU processing support through TPOT? That would allow for dramatically faster hyperparameter optimization.

rhiever:
I'm not sure what it would take to add XGBoost GPU support to TPOT. I'd imagine it entails that you can't set `n_jobs!=1`, but are there any additional contraints?

ap307:
In XGBoost you just have to able to pass the parameters `updater='grow_gpu_hist'` or `updater='grow_gpu'`.

[...]

PR #497, opened 2017-06-14
Rework n_jobs parameter

weixuanfu2016:
This PR is for solving issues #436 and #490 related to `joblib` by:

1. `joblib` module is not used in TPOT when n_jobs is equal to 1

2. Rework `_wrapped_cross_val_score` in TPOT based on [the source codes of `sklearn.model_selection.cross_val_score`](https://github.com/scikit-learn/scikit-learn/blob/14031f6/sklearn/model_selection/_validation.py#L36) and remove the `joblib` module in `cross_val_score` to avoid nested usage of `joblib` in TPOT


tpot/base.py
```diff
@@ -890,28 +890,31 @@ def _evaluate_individuals(self, individuals, features, target, sample_weight=Non
...
-                    max_eval_time_mins=self.max_eval_time_mins,
-                    groups=groups
-                )
-                jobs.append(job)
-            parallel = Parallel(n_jobs=self.n_jobs, verbose=0, pre_dispatch='2*n_jobs')
-            tmp_result_score = parallel(jobs)
-
+            if self.n_jobs == 1:
+                tmp_result_scores = [partial_wrapped_cross_val_score(sklearn_pipeline=sklearn_pipeline)
+                for sklearn_pipeline in sklearn_pipeline_list[chunk_idx:chunk_idx + self.n_jobs * 4]]
```
rhiever: This line should be indented more to make it clear that it's a continuation of the previous line.

rhiever: Also, there should be no chunking in the case where `n_jobs=1`, right?

tpot/base.py
```diff
@@ -890,37 +890,37 @@ def _evaluate_individuals(self, individuals, features, target, sample_weight=Non
...
-                                         'Continuing to the next pipeline.'.format(self._pbar.n))
-                    resulting_score_list.append(-float('inf'))
+        if self.n_jobs == 1:
+            for sklearn_pipeline in sklearn_pipeline_list:
+                val = partial_wrapped_cross_val_score(sklearn_pipeline=sklearn_pipeline)
+                resulting_score_list = self._update_pbar(val, resulting_score_list)
+        else:
+            # chunk size for pbar update
+            for chunk_idx in range(0, len(sklearn_pipeline_list), self.n_jobs * 4):
+                if self.n_jobs == 1:
```
rhiever: This check doesn't need to be here.

tpot/base.py
```diff
@@ -992,6 +992,19 @@ def _operator_count(self, individual):
                 operator_count += 1
         return operator_count
 
+    # update self._pbar during pipeline evaluration
```
rhiever: Please make this comment a docstring for the `_update_pbar` function.

weixuanfu2016: Okay, will do

Issue #499, opened 2017-06-15
fit() doesn't end properly (no best pipeline)


korcky:
fit() method don't even trying find best pipeline.

## Context of the issue

I'm trying to find parameters for my ensemble regressor, but when number of models for ensemble increase fit() method doesn't end properly - ending but don't even trying find best pipeline.

## Process to reproduce the issue

I'm using custom config dictionary, like this:
```
config_dict = {
    'ensemble.regressor.Ensemble': {
        'model1': {
            'ensemble.models.model1.Model1': {
                # parameters
            }
        }

        ...

        'model10': {
            'ensemble.models.model10.Model10': {
                # parameters
            }
        }
}
```
Then:
Step 1. `tpot = TPOTRegressor(..., config_dict=config_dict)`
Step 2. `tpot.fit(X_train, y_train)`
Step 3. `print(tpot.score(X_test, y_test))`

## Expected result

After step 2: best pipeline parameters
After step 3: score of best pipeline

## Current result

After step 2: nothing
After step 3:
```
RuntimeError                              Traceback (most recent call last)
<ipython-input-36-2ae990681f1a> in <module>()
----> 1 print(tpot.score(X_test, y_test))

/usr/local/lib/python3.5/dist-packages/tpot/base.py in score(self, testing_features, testing_target)
    660         """
    661         if self.fitted_pipeline_ is None:
--> 662             raise RuntimeError('A pipeline has not yet been optimized. Please call fit() first.')
    663 
    664         # If the scoring function is a string, we must adjust to use the sklearn

RuntimeError: A pipeline has not yet been optimized. Please call fit() first.
```

## Note

For this testing used dummy ensemble (simple regression where models of ensemble not used) and if delete all parameters of models (`# parameters`) in `config_dict`, or using only 2-3 models with parameters, then step 2 and 3 work correctly.
In tpot exist hidden max number of parameters or why this could happen?

weixuanfu2016:
A nice way to use TPOT. `config_dict` allows this kind of nested estimator as shown in our default [`config_dict`](https://github.com/rhiever/tpot/blob/master/tpot/config/regressor.py#L177-L185)  and we do not limit max number of parameters. Could you please use `verbosity=3` in `tpot = TPOTRegressor(..., config_dict=config_dict)` to check more detailed error message?

korcky:
Nice way to use such a nice thing. With `verbosity=3`:
```
[...]

PR #501, opened 2017-06-16
Removing "DEFAULT" value in terminals

weixuanfu2016:
1. Fix a bug that "DEFAULT" in the parameter(s) of nested estimator causes `KeyError` when exporting pipeline, which is related to the issue #499 

2. Add a unit test for this issue

The [2 lines in `operator_utils.py`](https://github.com/rhiever/tpot/blob/v0.8/tpot/operator_utils.py#L229-L230) cause the bug in the issue #499. It works all right with operators without nested/stacked estimator as parameter(s) but it will cause `KeyError` once `dep_op_arguments` do not have nested/stacked estimator's name as key of the dictionary. It means that if any parameter is `DEFAULT` in the nested/stacked estimator would cause the whole pipeline cannot exported as valid scikit-learn Pipeline object.

### DIFF
--- a/tests/export_tests.py
+++ b/tests/export_tests.py
@@ -335,7 +335,7 @@ def test_export_pipeline_5():
     tpot_obj = TPOTRegressor()
     pipeline_string = (
         'DecisionTreeRegressor(SelectFromModel(input_matrix, '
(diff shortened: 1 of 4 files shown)


weixuanfu2016:
Update: after discussions with @rhiever, we decide to remove `DEFAULT` value in terminal since it is confusing as default in TPOT. And user can easily use `config_dict` parameter to add default value back based on scikit-learn online documentation.

Issue #59, opened 2015-12-15
Smart seeding of TPOT populations?


rhiever:
Sorry if the text below sounds like rambling -- I was using this issue to brainstorm.

I've been thinking about possible ways to make TPOT perform better right out of the box, without having to run it for several generations to finally discover the better pipelines. One of the ideas I've had is to seed the TPOT population with a smarter group of solutions.

For example, we know that a TPOT pipeline will need at least one model, so we can seed it with each of the 6 current models over a small range of parameters:

* decision tree: all combinations of
    * max_features = [0 (--> auto), 1 (--> None)]
￼￼￼￼￼￼￼￼￼    * max_depth: [0 (--> None), 1, 5, 10, 20, 50]
    * = 12 total combinations

* random forest: all combinations of
    * n_estimators = [100, 500]
    * max_features = [0 (--> auto), 1]
    * = 4 total combinations

* logistic regression:
    * C = [0.01, 0.1, 0.5, 1.0, 10.0, 50.0, 100.0]
    * = 7 total combinations

* svc:
    * C = [0.01, 0.1, 0.5, 1.0, 10.0, 50.0, 100.0]
    * = 7 total combinations

* knnc:
    * n_neighbors = [2, 5, 10, 20, 50]
    * = 5 total combinations

* gradient boosting: all combinations of
    * learning_rate: [0.01, 0.1, 0.5, 1.0, 10.0, 50.0, 100.0]
    * n_estimators: [100, 500]
￼￼￼￼￼￼￼￼￼    * max_depth: [0 (--> None), 1, 5, 10, 20, 50]
    * = 84 total combinations

That gives us 119 "classifier-only" TPOT pipelines to start with.

We also have 4 feature selectors:

* RFE: all combinations of
    * num_features = [1, 5, 10, 50]
    * step = [0.1, 0.25, 0.5]
    * = 12 total combinations

* select percentile:
    * percentile = [1, 5, 10, 25, 50, 75]
    * = 6 total combinations

* select k best:
    * k = [1, 2, 5, 10, 20, 50]
    * = 6 total combinations

* variance threshold:
    * threshold = [0.1, 0.2, 0.3, 0.4, 0.5]
    * = 5 total combinations

And 4 feature preprocessors:

* standard scaler (no parameters)
    * = 1 total combinations

* robust scaler (no parameters)
    * = 1 total combinations

* polynomial features (no parameters)
    * = 1 total combinations

* PCA:
[...]

PR #502, opened 2017-06-21
Add capability to provide custom seeds to GP

teaearlgraycold:
## What does this PR do?

* Adds a `seeds` parameter to `TPOTBase`
* Allows for users to export a list named `seeds` in their custom config

**Customizing TPOT's starting population** in the "Using" section of the docs.

In this  function in `TPOTBase`:

```python
def _setup_pop(self, seeds, config_path):
```

There are tests added to the `tpot_tests.py` test file.

#59 
#296


tpot/base.py
```diff
@@ -355,23 +358,36 @@ def _read_config_file(self, config_path):
...
+            if hasattr(config, 'seeds'):
+                seeds = config.seeds
+        except Exception as e:
+            # Config isn't a file
+            return
+
+        seed_individuals = [creator.Individual.from_string(x, self._pset) for x in seeds]
+        self._pop = []
+
+        # Add the same set of seeds to the population until we have population_size seeds
```
weixuanfu2016: With a little discussions with @rhiever earlier today, we think the `seeds` (maybe need better name for this parameter) should only specify a small part of initial population instead of initial population that is full of duplicated pipelines in the `seeds` (If `seeds` is a small list, then the pipeline diversity of GP will be limited in the beginning).  Other pipelines in initial population rather than pipelines in `seeds` should be randomly generated.

rhiever: `population_seeds` should probably be the parameter name.

docs/using/index.html
```diff
@@ -576,6 +571,42 @@ <h1 id="customizing-tpots-operators-and-parameters">Customizing TPOT's operators
...
 <p>Note that you must have all of the corresponding packages for the operators installed on your computer, otherwise TPOT will not be able to use them. For example, if XGBoost is not installed on your computer, then TPOT will simply not import nor use XGBoost in the pipelines it considers.</p>
+<h1 id="customizing-tpots-starting-population">Customizing TPOT's starting population</h1>
+<p>TPOT allows for the initial population of pipelines to be seeded. This can be done either through the <code>population_seeds</code> parameter in the TPOT constructor, or through a <code>population_seeds</code> attribute in a custom config file.</p>
+<pre><code class="Python">population_seeds = [
+    'BernoulliNB(GaussianNB(input_matrix), BernoulliNB__alpha=0.1, BernoulliNB__fit_prior=False)',
+    'BernoulliNB(input_matrix, BernoulliNB__alpha=0.01, BernoulliNB__fit_prior=True)'
```
rhiever: How easy would it be to take actual sklearn pipelines as input instead of the string representations? I sense that users would find it easier to seed with sklearn pipelines rather than this (slightly) weird string representation we use.

rhiever: Also, we should use better example clfs in the docs: Maybe a RandomForestClassifier and a LogisticRegression?

tpot/base.py
```diff
@@ -1068,7 +1081,7 @@ def _operator_count(self, individual):
         return operator_count
 
     def _update_pbar(self, val, resulting_score_list):
-        """Update self._pbar during pipeline evaluration
+        """Update self._pbar during pipeline evaluration.
```
rhiever: Not your typo, but there's a typo here in "evaluation."

tests/tpot_tests.py
```diff
@@ -297,6 +297,22 @@ def test_conf_dict_3():
...
+    assert isinstance(tpot_obj._pop, list)
+    assert isinstance(tpot_obj._pop[0], creator.Individual)
+
+
+def test_conf_dict_5():
+    """Assert that TPOT has a _pop attribute of the same size as the number of seeds provided."""
+    tpot_obj = TPOTRegressor(config_dict='tests/test_config.py')
+    n_seeds = len(tpot_obj._read_config_file('tests/test_config.py').population_seeds)
+
+    assert len(tpot_obj._pop) == n_seeds
```
rhiever: I believe the unit tests need to be updated.

docs/using/index.html
teaearlgraycold: >How easy would it be to take actual sklearn pipelines as input instead of the string representations?

We don't have any code to go from sklearn pipelines to deap pipelines, only the other direction. It would be non-trivial to write, but doable given Python's complete reflection of objects.

>we should use better example clfs in the docs: Maybe a RandomForestClassifier and a LogisticRegression?

Sure, that's an easy enough change.

tests/tpot_tests.py
teaearlgraycold: There seems to be just one that broke, and only in 2.7. I'll install conda for 2.7 so I can see what went wrong.

docs/using/index.html
rhiever: It would be nice to have a function to go from sklearn pipelines to DEAP pipelines. :-) That will be one step closer to having DEAP work directly on sklearn pipelines themselves.

teaearlgraycold: Alright. I'll start working on that.

rhiever: How do you propose doing that? We need to make sure that it only takes in the parameters that are defined in the config.

teaearlgraycold: Pass in the config that's being used and check that the pipeline is valid (within the context of the config) as you walk though it.

rhiever: Sounds good.

weixuanfu2016: @teaearlgraycold I think the function to go from sklearn pipeline to DEAP pipelines was not added in this PR yet.  But I will start merging some PRs to dev branch today in case of tons of conflicts between PRs. Please add this function later in another PR.

Issue #503, opened 2017-06-21
Tpot examples do not seem to differentiate/evolve?


dartdog:
Running the example the initial optimization values found do not change?

## Context of the issue

so for instance on the Iris classifier example I ran it and got the following: 

/home/tom/anaconda3/envs/py36n/lib/python3.6/site-packages/sklearn/cross_validation.py:44: DeprecationWarning: This module was deprecated in version 0.18 in favor of the model_selection module into which all the refactored classes and functions are moved. Also note that the interface of the new CV iterators are different from that of this module. This module will be removed in 0.20.
  "This module will be removed in 0.20.", DeprecationWarning)
Optimization Progress:  31%|███       | 92/300 [00:23<00:32,  6.41pipeline/s]
Generation 1 - Current best internal CV score: 0.9746376811594203
Optimization Progress:  47%|████▋     | 141/300 [00:39<00:30,  5.23pipeline/s]
Generation 2 - Current best internal CV score: 0.9746376811594203
Optimization Progress:  63%|██████▎   | 190/300 [00:50<00:14,  7.69pipeline/s]
Generation 3 - Current best internal CV score: 0.9746376811594203
Optimization Progress:  77%|███████▋  | 231/300 [00:57<00:07,  8.68pipeline/s]
Generation 4 - Current best internal CV score: 0.9746376811594203
                                                                              
Generation 5 - Current best internal CV score: 0.9746376811594203

Best pipeline: GaussianNB(input_matrix)
0.921052631579

So best internal CV score Never changes..

Seeing the same when running the Minst example.. IE no change in scores..

One would expect to see some search activity with differing values??

minst output:

Optimization Progress:  31%|███▏      | 94/300 [05:29<04:56,  1.44s/pipeline]
Generation 1 - Current best internal CV score: 0.9859273244941604
Optimization Progress:  48%|████▊     | 143/300 [10:57<12:22,  4.73s/pipeline]
Generation 2 - Current best internal CV score: 0.9859273244941604
Optimization Progress:  62%|██████▏   | 187/300 [13:09<04:19,  2.30s/pipeline]
Generation 3 - Current best internal CV score: 0.9859273244941604
Optimization Progress:  77%|███████▋  | 231/300 [15:04<03:09,  2.74s/pipeline]
Generation 4 - Current best internal CV score: 0.9859273244941604
                                                                              
Generation 5 - Current best internal CV score: 0.9859273244941604

[...]

PR #506, opened 2017-06-23
Reduce the number of repeated pipelines between generations

PG-TUe:
During the TPOT optimization process, many populations across generations often had many duplicates because of some bugs. This lead to much fewer unique pipelines being evaluated than expected.

## What does this PR do?
It fixes said bugs, now there will *almost* never be duplicates in the generation, nor individuals which have been evaluated in earlier generations.
I did a trial run on the iris dataset, with 100 individuals for 36 generations, on average only 1.83 of the population in a generation was duplicate, and 0.33 had a 'cache hit' (meaning they already were in `evaluated_individuals` before the generation).

### [Crossover](https://github.com/PG-TUe/tpot/blob/no_repeats/tpot/base.py#L1024-L1032)
Nothing too crazy, instead of calling crossover once, it is called repeatedly until an offspring which has not been evaluated before is found, or for at most 50 times (arbitrary stopping point).
Additionally, I made sure both individuals passed to this function had at least two Primitives each [here](https://github.com/PG-TUe/tpot/blob/no_repeats/tpot/gp_deap.py#L78-L84). 
This is not a perfect solution, see #503, however it is important because often two individuals which could not generate new offspring would be passed to crossover (each having only one primitive of a different type).

### [Mutation](https://github.com/PG-TUe/tpot/blob/no_repeats/tpot/base.py#L1034-L1074)
Similar to above, if the operator does not generate a new individual, it is tried again, up to 50 times.
Additionally, there is now a check to make sure that individuals with only one primitive will not be mutated with the `mutShrink` operation, since they can not shrink (which previously resulted in the exact same individual being returned).
Finally, it can be the case that the number of shrink options is quite limited, eg. an individual with only two primitives can have only one primitive shrunk, this can lead to a scenario where an individual as well as all of its shrunk versions have already been evaluated.
If this is the case (here this is determined if the `mutShrink` operator is chosen and does not return an unseen individual after 50 tries - it is not determined exhaustively), then the individual is again passed to the function, but without allowing the `mutShrink` mutator to be chosen, since a different operator is likely to give a new individual.

One can inspect the number of duplicates in a population by keeping track of them, for example add `print('duplicates this generation:',len(individuals) - len(unique_individuals))` [here](https://github.com/PG-TUe/tpot/blob/no_repeats/tpot/base.py#L940).
One can inspect the number of cache hits by adding a counter [here](https://github.com/PG-TUe/tpot/blob/no_repeats/tpot/base.py#L960).

All in #503.

## Remark
[This commit](https://github.com/rhiever/tpot/commit/c45c3002a45f6fd73836a0bd5c771252f30abc34) of this branch also adds a small code snippet which allows individuals to be formatted as string without the ugly verbose parameter prefixes [here](https://github.com/PG-TUe/tpot/blob/no_repeats/tpot/base.py#L734-L755). This is now used when presenting the best found pipeline to the user [here](https://github.com/PG-TUe/tpot/blob/no_repeats/tpot/base.py#L603-L604).
For example 'BernoulliNB(input_matrix, BernoulliNB__alpha=1.0, BernoulliNB__fit_prior=True)' would be formatted as 'BernoulliNB(input_matrix, alpha=1.0, fit_prior=True)'.

Finally, again the same one nosetest fails for me (below), but other than that it seems passes all on my end.
`ERROR: Assert that ZeroCount operator returns correct transformed X.`


PG-TUe:
I noticed that the development branch no longer has `tests.py`, how should I submit a new test for my code?

Here is a code which covers the clean_pipeline_string function:
```
def test_clean_pipeline_string():
    """Assert that clean_pipeline_string correctly returns a string without parameter prefixes"""

    with_prefix = 'BernoulliNB(input_matrix, BernoulliNB__alpha=1.0, BernoulliNB__fit_prior=True)'
    without_prefix = 'BernoulliNB(input_matrix, alpha=1.0, fit_prior=True)'
    tpot_obj = TPOTClassifier()
    ind1 = creator.Individual.from_string(with_prefix)
    
    pretty_string = tpot_obj.clean_pipeline_string(ind1)
    assert pretty_string == without_prefix
```

weixuanfu2016:
@PG-TUe check tpot_tests.py in tests folder instead.

rhiever:
This PR looks promising and I'm generally supportive of what it implements. I will try to review it early next week so we can get it merged.

PG-TUe:
Okay, I looked at the coverage, I am not sure why [line 85-93](https://coveralls.io/builds/12109785/source?filename=tpot%2Fgp_deap.py#L85) is not covered if 84 is (only reason would be a crash? but then this should follow from test results?).

I will add unit tests for mutate and mate operators.

rhiever:
That's a strange one. You can drop a debugger at that line and run the specific unit test for that logical branch to see what's going on. A crash seems to be the logical explanation. I think our unit tests sometimes squash the errors (though they shouldn't).

PG-TUe:
I checked it out. It was indeed a crash caused by a population consisting of only individuals with one primitive. What would be the desired behavior here? Especially when working with a small population it could be realistic that this comes up in practice.

I think I should be able to create a new selection procedure which always selects two individuals that can perform crossover, even if they have only one primitive (the same one). However, I do not have the time to do so now or tomorrow. I might be able to do this monday or tuesday. The simple approach to do this would be:
(1) select a random individual
(2) find all primitives in this individual
(3) find all other individuals in the population with at least one matching primitive\*
(4) select a random individual from those individuals
(5) perform crossover
This should be linear time in the total length of all pipelines, so I think that is quite manageable.
\* Technically should be one matching terminal, but the only terminal that exists that would match between individuals that do not share a primitive would be the `input_matrix` and only swapping those out does not create a new individual.

That said, if there is a population in which no individual shares any primitives, this still would crash.
In practice this should probably not come up as (1) populations are quite big, so duplicate primitives should be there and (2) pipelines would probably always tend to use a similar set of algorithms that perform well for a certain dataset.
If we do find such an odd population, we can either (1) allow crossover anyway (switching input matrices, doing nothing) or (2) disable crossover and only allow mutation.
I prefer option two because is more likely to introduce individuals with shared primitives for the next generation (but again, this probably does not come up in practice).

PG-TUe:
Just to be clear, commits above are just refactoring, not additional functionality discussed.

PG-TUe:
Okay, changes are made and unit tests added. I also merged what was needed after PR #510 was made (not much). Crossover now works between any two individuals that are distinct from each other but share at least one primitive.

Only thing that would need clarification is the desired behavior if:
* the population is of size 0 or 1.
* there are no shared primitives between any two individuals in the population.

The first should not happen, the second is very unlikely but could happen.

tpot/gp_deap.py
```diff
@@ -38,8 +38,37 @@
...
+    eligible_pairs = [(i, i+1+j) for i, ind1_prims in enumerate(primitives_by_ind)
+                                 for j, ind2_prims in enumerate(primitives_by_ind[i+1:])
+                                 if not ind1_prims.isdisjoint(ind2_prims) and
+                                    pop_as_str[i] != pop_as_str[i+1+j]]
+
+    eligible_pairs += [(j, i) for (i,j) in eligible_pairs]
+    
+    if not eligible_pairs:
+        # what TODO? placeholder
+        return population[0], population[0]
```
rhiever: If there are no eligible pairs, then I think your current solution is fine. As you mentioned, it's not ideal because it's essentially a copy operation.

We could have it return `None, None` in this case and have the calling code know to interpret that as a "use mutation instead" signal. If we encapsulate the mutation code into a function, maybe that won't be so bad.

tpot/base.py
```diff
@@ -632,7 +633,9 @@ def _update_top_pipeline(self, features, target):
                     # Add an extra line of spacing if the progress bar was used
                     if self.verbosity >= 2:
                         print('')
-                    print('Best pipeline: {}'.format(self._optimized_pipeline))
+                    
+                    optimized_pipeline_str = self.clean_pipeline_string(self._optimized_pipeline)
+                    print('Best pipeline:', optimized_pipeline_str)
```
rhiever: Minor issue: Will this still have a space between `Best pipeline:` and the pipeline string?

tpot/base.py
```diff
@@ -1018,10 +1046,16 @@ def _evaluate_individuals(self, individuals, features, target, sample_weight=Non
...
+            offspring, _ = cxOnePoint(ind1_copy, ind2_copy)
+            if str(offspring) not in self.evaluated_individuals_:
+                break
+        
+        return offspring, _
 
     @_pre_test
-    def _random_mutation_operator(self, individual):
+    def _random_mutation_operator(self, individual, allow_shrink=True):
         """Perform a replacement, insertion, or shrink mutation on an individual.
```
rhiever: Please document the `allow_shrink` parameter in the docstrings.

tpot/base.py
```diff
@@ -85,6 +85,7 @@ class TPOTBase(BaseEstimator):
 
     # dont save periodic pipelines more often than this
     OUTPUT_BEST_PIPELINE_PERIOD_SECONDS = 30
+    MAX_MUT_LOOPS = 50
```
rhiever: Please refactor both of these constants to internal variables (with a `_` in front).

I know your PR doesn't add `OUTPUT_BEST_PIPELINE_PERIOD_SECONDS`, but might as well fix both here.

tests/tpot_tests.py
```diff
@@ -823,3 +883,14 @@ def test_gen():
...
+def test_clean_pipeline_string():
+    """Assert that clean_pipeline_string correctly returns a string without parameter prefixes"""
+
+    with_prefix = 'BernoulliNB(input_matrix, BernoulliNB__alpha=1.0, BernoulliNB__fit_prior=True)'
+    without_prefix = 'BernoulliNB(input_matrix, alpha=1.0, fit_prior=True)'
+    tpot_obj = TPOTClassifier()
+    ind1 = creator.Individual.from_string(with_prefix, tpot_obj._pset)
+    
+    pretty_string = tpot_obj.clean_pipeline_string(ind1)
+    assert pretty_string == without_prefix
```
rhiever: Please add a newline at the end of the file here.

tpot/base.py
```diff
@@ -1018,10 +1046,16 @@ def _evaluate_individuals(self, individuals, features, target, sample_weight=Non
 
     @_pre_test
     def _mate_operator(self, ind1, ind2):
-        return cxOnePoint(ind1, ind2)
+        for __ in range(self.MAX_MUT_LOOPS):
```
rhiever: Only one underscore should be needed?

tpot/gp_deap.py
```diff
@@ -75,32 +104,22 @@ def varOr(population, toolbox, lambda_, cxpb, mutpb):
...
-            if ind_str != str(ind):  # check if mutation happened
-                del ind.fitness.values
+            idx = np.random.choice(len(population))
+            ind = population[idx]
+            ind, = toolbox.mutate(ind)            
+            del ind.fitness.values
             offspring.append(ind)
         else:  # Apply reproduction
-            idx = np.random.randint(0, len(population))
+            idx = np.random.choice(len(population))
```
rhiever: It might be preferable to use `randint` here because it's more explicit. It wasn't immediately obvious that `choice` was correct here until I looked at the docs to confirm that `np.random.randint(0, len(population))` == `np.random.choice(len(population))`.

tpot/base.py
```diff
@@ -751,6 +754,31 @@ def set_params(self, **params):
 
         return self
 
+    def clean_pipeline_string(self, individual):
+        """
```
rhiever: Minor issue: Please remove the newline between the `"""` and the start of the docstring text.

tpot/gp_deap.py
```diff
@@ -38,8 +38,37 @@
 import warnings
 import threading
 
-# Limit loops to generate a different individual by crossover/mutation
-MAX_MUT_LOOPS = 50
+def pick_two_individuals_eligible_for_crossover(population):
+    """ Pick two individuals from the population which can do crossover, that is, they share a primitive.
```
rhiever: Minor issue: Please remove the space between the """ and the start of the docstring text.

tpot/gp_deap.py
```diff
@@ -38,8 +38,37 @@
...
+    primitives_by_ind = [set([node.name for node in ind if isinstance(node, gp.Primitive)])
+                         for ind in population]
+    pop_as_str = [str(ind) for ind in population]
+    
+    eligible_pairs = [(i, i+1+j) for i, ind1_prims in enumerate(primitives_by_ind)
+                                 for j, ind2_prims in enumerate(primitives_by_ind[i+1:])
+                                 if not ind1_prims.isdisjoint(ind2_prims) and
+                                    pop_as_str[i] != pop_as_str[i+1+j]]
+
+    eligible_pairs += [(j, i) for (i,j) in eligible_pairs]
```
rhiever: Please add a comment describing that this is ensuring that eligible pairs in both orders (i, j) and (j, i) are considered.

rhiever:
I made some (mostly) minor comments about the PR. Otherwise, this PR LGTM. 👍 

Here's a broader issue that came to mind while reviewing this PR: Could forcing the GP to always produce unexplored individuals from mutation/crossover hamper the GP algorithm?

For example, let's say the GP finds a nice pipeline with 3 primitives that performs well. It may make some variants of that pipeline---say, a couple offspring that add 1 primitive to the pipeline each.

However, what if the best pipeline is actually a version of the 3-primitive pipeline with 1 of the primitives removed? Mutation/crossover won't be able to change those pipelines back to the original 3-primitive pipeline, so the GP will be challenged to find a way to jump to the ideal 2-primitive pipeline from the 4-primitive pipeline variants.

Basically: It's probable that having copies of already-evaluated pipelines in the population---especially good pipelines---can be useful as material for mutations and crossover to work on. I think before we merge this PR (and after we settle the code), we'll have to do some thorough comparisons between TPOT with and without this change to make sure that it doesn't hamper TPOT's abilities to explore the pipeline space.

tpot/base.py
PG-TUe: Yes, in both [python 2](https://docs.python.org/2/library/functions.html#print) and [python 3](https://docs.python.org/3/library/functions.html#print), a blank space (' ') is the default value for the `sep` parameter which indicates how the strings should be separated in a print function.

rhiever:
I discussed the above issue with some folks in the lab and it may be less of a concern for TPOT because it uses the mu+lambda algorithm. In the example I used above, the 3-primitive pipeline would not be replaced on the Pareto front by the 4-primitive variants because it is less complex. So the selection procedure that eliminates half of the population would still keep the 3-primitive pipeline around, as long as there isn't another 3-primitive pipeline that outperforms it.

tpot/base.py
PG-TUe: Hm yeah, I used two so it did not have the same name as the second return argument from the crossover function, because that value is also not used. I wasn't quite sure what to do here, I played around a bit to see what pylint recommended, but nothing came of that either.
I'll change it back to one.

PG-TUe:
Yes, that should be exactly correct. Pareto-front selection, in addition to the fact that besides the offspring the previous generation is always considered, should handle that case.

tpot/gp_deap.py
PG-TUe: I agree (I initially changed it so it would match the function used when selecting a pair for crossover). 
I should point out that `np.random.randint(0, len(population))` == `np.random.randint(len(population))`, would you prefer the explicit `0` back?

tpot/gp_deap.py
PG-TUe: I will change this, I agree the use of `None` would be more appropriate.

PR #515, opened 2017-06-28
More unit tests

weixuanfu2016:
More unit tests for driver and a small unit test for `_warpped_cross_eval_score` to increase code coverage.

Update: fix a issue about no  FileNotFoundError in Python 2.*


tpot/driver.py
```diff
@@ -442,18 +442,17 @@ def load_scoring_function(scoring_func):
...
             print('taken from module: {}'.format(module_name))
         except Exception as e:
             print('failed importing custom scoring function, error: {}'.format(str(e)))
             raise ValueError(e)
-    
+
     return scoring_func
 
-def main():
+def main(args):
```
rhiever: This change will break the CLI.

weixuanfu2016: CLI issue is fixed in the commits below

tests/test_config.py.bad
```diff
@@ -0,0 +1,35 @@
+# -*- coding: utf-8 -*-
```
rhiever: Please rename this file to `test_config_bad.py`.

weixuanfu2016: That would cause a error during unit tests since it is not in right format as python codes. Check this [link](https://ci.appveyor.com/project/rhiever/tpot/build/1.0.79/job/wuosvq3apiysux7a)

rhiever: That's confusing. The CI systems are running that file?

weixuanfu2016: `nosetests -s -v` would check all the .py files in tests folder.

Issue #536, opened 2017-07-26
Feature Selection scoring using `f_classif` in both `regressor.py` and `regressor_light.py`


dclambert:
## Context of the issue

The standard and light configs for regression use `sklearn.feature_selection.f_classif` as the score function for both `SelectPercentile` and `SelectFwe` rather than `sklearn.feature_selection.f_regression`.  

Is this intentional?

weixuanfu:
Hmm, I think it is a bug. I will submit a PR for fixing it.

weixuanfu:
@dclambert Thank you for reporting this issue.

I just posted a PR #538 for it and it is based on dev branch. You may try the commands below to reinstall TPOT with the patch

```
pip install --upgrade --no-deps --force-reinstall git+https://github.com/weixuanfu/tpot.git@f_regression
```

Also I made another branch with the patch based on Master branch (I did not post PR for it for avoiding conflicts). Try the commands below for installing the branch. 

```
pip install --upgrade --no-deps --force-reinstall git+https://github.com/weixuanfu/tpot.git@f_regression_patch
```

PR #538, opened 2017-07-26
Use f_regression for feature selection in regression tasks

weixuanfu:
## What does this PR do?

Use `sklearn.feature_selection.f_regression` instead of `sklearn.feature_selection.f_classif` for regression tasks.

Related issue #536

### DIFF
--- a/tpot/operator_utils.py
+++ b/tpot/operator_utils.py
@@ -243,7 +243,7 @@ def TPOTOperatorClassFactory(opsourse, opdict, BaseClass=Operator, ArgBaseClass=
                 # To make sure the inital operators is the first parameter just
                 # for better persentation
                 for dep_op_pname, dep_op_str in dep_op_list.items():
-                    if dep_op_str == 'f_classif':
+                    if dep_op_str in ['f_classif', 'f_regression']:
                         arg_value = dep_op_str
                     else:
                         arg_value = "{}({})".format(dep_op_str, ", ".join(dep_op_arguments[dep_op_str]))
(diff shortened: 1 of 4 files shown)


tpot/operator_utils.py
```diff
@@ -243,7 +243,7 @@ def export(cls, *args):
                 # To make sure the inital operators is the first parameter just
                 # for better persentation
                 for dep_op_pname, dep_op_str in dep_op_list.items():
-                    if dep_op_str == 'f_classif':
+                    if dep_op_str in ['f_classif', 'f_regression']:
```
rhiever: Why is this check needed?

weixuanfu: Because `SelecPercentile` do not take `f_classif()` or `f_regression()`
```
In [3]: selector = SelectPercentile(f_classif, percentile=10)
   ...: selector.fit(X, y)
   ...:
Out[3]:
SelectPercentile(percentile=10,
         score_func=<function f_classif at 0x000002A3C66CF6A8>)

In [4]: selector = SelectPercentile(f_classif(), percentile=10)
   ...: selector.fit(X, y)
   ...:
   ...:
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-4-bba9fbc03e91> in <module>()
----> 1 selector = SelectPercentile(f_classif(), percentile=10)
      2 selector.fit(X, y)
      3

TypeError: f_classif() missing 2 required positional arguments: 'X' and 'y'
```

rhiever: Is there a more general way that we can handle this situation? Programming in specific function names doesn't seem to be ideal. We'll run into the same issue with SelectPercentile for other scoring functions as well.

weixuanfu: Check this [commit](https://github.com/rhiever/tpot/pull/538/commits/ea3b1945847f2ce5498c456a1a98c6dcdd4b2e3f) that is a more general way since `score_func` in scikit-learn only gets callable function.

rhiever: That seems like a better solution for now. 👍

Issue #539, opened 2017-07-30
Is it possible to add "early stopping" in case the accuracy pass a specific value.


drorhilman:
Question: 
TPOT runs can be very long for large population, but sometime you reach a very high accuracy early-on, and you would like the process to stop if it is not improving after a while. is it possible to do it? 
For example if it reaches 99% accuracy and not improving in 3 generation then stop.

weixuanfu:
Hmm, I think we might add this option once the accuracy of classification reaches a very high score, like 0.999. What do you think of this idea? @rhiever 

But I think it is not simply to use the option in most of scoring functions in regression tasks.

drorhilman:
Yes. I think that it should have a notion of "progress" - check whether you are still improving or not, and if not then stop.

weixuanfu:
If the option just check whether the best pipeline is still improving in GP, this option maybe misused with small number of population size due to limited number of new better individuals per generation. I think we still need a threshold value if we add this option.

weixuanfu:
@drorhilman after some discussions with @rhiever, we will add this option to stop process if the score does not improve after N generations. The option will be added into next version of TPOT and of course, it is turned off by default.

weixuanfu:
Update: We added this option in the PR #541 and it should be merged later.

PR #541, opened 2017-08-04
Add "early stopping" option for stopping optimization process

weixuanfu:
## What does this PR do?

1. refine `self._check_periodic_pipeline`

2. Add "early stopping" option for stopping optimization process when no improvement in a set number of generations

[base.py](https://github.com/weixuanfu/tpot/blob/ca19ca479cc2b68ce8f54ef97ea23955728eadeb/tpot/base.py#L837)

```
import numpy as np

iris = load_iris()
    iris.target.astype(np.float64), train_size=0.75, test_size=0.25, random_state=42)

tpot = TPOTClassifier(generations=20, population_size=10, verbosity=2, early_stop=2, random_state=42)
tpot.fit(X_train, y_train)
```

#539 

## Questions:

- Do the docs need to be updated? Yes, docs were updated in the PR

### DIFF
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -237,8 +237,8 @@ class TPOTBase(BaseEstimator):
         self.max_eval_time_seconds = max(int(self.max_eval_time_mins * 60), 1)
         self.periodic_checkpoint_folder = periodic_checkpoint_folder
         self.early_stop = early_stop
-        self._last_optimized_pipeline = None
-        self._last_optimized_pipeline_n_gens = 0
+        self._last_optimized_pareto_front = None
+        self._last_optimized_pareto_front_n_gens = 0
 
         # dont save periodic pipelines more often than this
         self._output_best_pipeline_period_seconds = 30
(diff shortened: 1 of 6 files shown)


tpot/base.py
```diff
@@ -636,32 +631,66 @@ def _update_top_pipeline(self, features, target):
                                   'TPOTClassifier object. Please make sure you '
                                   'passed the data to TPOT correctly.')
             else:
-                self.fitted_pipeline_ = self._toolbox.compile(expr=self._optimized_pipeline)
+                if not self._last_optimized_pipeline:
+                    self._last_optimized_pipeline = str(self._optimized_pipeline)
+                elif self._last_optimized_pipeline == str(self._optimized_pipeline):
```
rhiever: Maybe the check should be whether the Pareto front has changed at all in the last *n* generations, rather than whether the single "best" pipeline has changed.

tpot/base.py
```diff
@@ -805,30 +834,34 @@ def clean_pipeline_string(self, individual):
...
+        if self.periodic_checkpoint_folder is not None:
+            total_since_last_pipeline_save = (datetime.now() - self._last_pipeline_write).total_seconds()
+            if total_since_last_pipeline_save > self._output_best_pipeline_period_seconds:
+                self._last_pipeline_write = datetime.now()
+                self._save_periodic_pipeline()
 
+        if self.early_stop is not None:
+            if self._last_optimized_pipeline_n_gens >= self.early_stop:
+                raise StopIteration("The optimized pipeline was not improved after evaluating {} more generations. "
+                                        "Will kill the optimization process.\n".format(self.early_stop))
```
weixuanfu: OK, make sense :-P

tpot/base.py
weixuanfu: Hmm, I need test it before using Pareto front.

weixuanfu: Fixed~

PR #552, opened 2017-08-28
OHE auto feature selection fix

### LINKED ISSUE #525: [dev] OneHotEncoder bug?
I've been using the OneHotEncoder that we ported from auto-sklearn for a separate project and keep getting an error message when using it in a pipeline.

Here's a minimal example that reproduces the issue:

```python
from sklearn.linear_model import LinearRegression
from sklearn.datasets import load_boston
from sklearn.pipeline import make_pipeline
from sklearn.model_selection import cross_val_score, KFold
from tpot.builtins import OneHotEncoder

boston = load_boston()

clf = make_pipeline(OneHotEncoder(sparse=False, minimum_fraction=0.05), LinearRegression())
cross_val_score(clf, boston.data, boston.target, cv=KFold(n_splits=10, shuffle=True))
```

>ValueError: X has different shape than during fitting. Expected 2, got 3.

Oddly, this doesn't happen with lower `n_splits`. It's also not limited to classification problems.

Any ideas what's going on? My initial guess is that it's not resetting for each fold, i.e., it learns the transformation on one fold and tries to use the same transformation on other folds.

cc @teaearlgraycold @mfeurer

danthedaniel:
Fixes #525 and adds a new regression test.

### DIFF
--- a/tests/one_hot_encoder_tests.py
+++ b/tests/one_hot_encoder_tests.py
@@ -31,7 +31,10 @@ SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
 import numpy as np
 import scipy.sparse
 from sklearn.utils.testing import assert_array_almost_equal
-from sklearn.datasets import load_iris
+from sklearn.datasets import load_iris, load_boston
+from sklearn.linear_model import LinearRegression
+from sklearn.pipeline import make_pipeline
+from sklearn.model_selection import cross_val_score, KFold
 from nose.tools import assert_equal
 
 from tpot.builtins.one_hot_encoder import OneHotEncoder, _auto_select_categorical_features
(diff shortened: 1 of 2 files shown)


tests/one_hot_encoder_tests.py
```diff
@@ -276,3 +279,18 @@ def test_transform_with_unknown_value():
     tds = scipy.sparse.csr_matrix(test_data)
     output = ohe.transform(tds).todense()
     assert np.sum(output) == 3
+
+
+def test_k_fold_cv():
```
weixuanfu, line 284: Maybe need a simple docstring for each unit test in one hot encoder.

tpot/builtins/one_hot_encoder.py
```diff
@@ -372,6 +370,9 @@ def fit_transform(self, X, y=None):
         y: array-like {n_samples,} (Optional, ignored)
             Feature labels
         """
+        if self.categorical_features == "auto":
+            self.categorical_features = _auto_select_categorical_features(X)
```
weixuanfu, line 374: Nice fix!

tpot/builtins/one_hot_encoder.py
```diff
@@ -147,7 +144,8 @@ class OneHotEncoder(BaseEstimator, TransformerMixin):
     categorical_features: "all" or array of indices or mask
         Specify what features are treated as categorical.
 
-        - 'all' (default): All features are treated as categorical.
+        - 'all': All features are treated as categorical.
+        - 'auto' (default): Select only features that have less than 10 unique values.
```
rhiever, line 148: I think the # of unique values for the auto selection should be parameterized for the OneHotEncoder class, with a default of 10.

Issue #549, opened 2017-08-16
Add feature selection operators to select categorical or continuous features


rhiever:
Currently, TPOT operators act on the full dataset regardless of the feature types. For example, if TPOT decides to apply a OneHotEncoder on the data, then the OneHotEncoder will be applied to all features---even continuous features. This issue can be problematic for datasets with mixed feature types.

We should look into adding feature selection operators that select *only continuous* or *only categorical* features. These operators will allow TPOT to subset the features by feature *type* and (hopefully) further apply appropriate transformations to those features.

An example pipeline I can imagine with these operators is, e.g.,

input --> select continuous --> PCA ----------------
&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;|-----> RandomForest
input --> select categorical --> OneHotEncoder ---

This should be a simple addition to the current version of TPOT, as all it requires is the addition of two new built-in feature selection operators. We could have the operators decide that a feature is categorical if it has < 10 levels, and it is continuous otherwise. We can make the threshold value a parameter for these operators, with a default of 10.

JuanuMusic:
Another useful way is to detect categorical types on pandas dataframes. Of course, this should go along with a good documentation about the behaviour.

Also, there could be a parameter indicating column indices or names of categorical values features.

lstmemery:
[...]

PR #560, opened 2017-09-06
Add feature selection operators to select categorical or continuous features

weixuanfu:

## What does this PR do?

For the issue #549:

1. Add too builtin operators/transfomer,  CategoricalSelector and ContinuousSelector, for feature selection

2. refine OneHotEncoder codes

3. Add unit tests for them

builtins/ feature_transformers.py

Any TPOT examples

#549 

## Questions:


bollwyvl:
Hi folks, love the work.

I've been mainly focused on the build side of these things to support other developers, but am now scaling up my own understanding of the sklearniverse beyond "will the tests run". I like tpot because it already wraps a lot of cool things, which i can try out and play against each other.

Open PRs are always interesting, so I've been digging into this one. While I haven't tried it out locally, this one looks good, as I've certainly seen datasets that mix continuous and categorical data! 

A quick thing I notice is that some of the categorical encoders (plus a whole raft more) are already implemented over in [category_encoders](https://github.com/scikit-learn-contrib/categorical-encoding), which is already over on [conda-forge](https://github.com/conda-forge/category_encoders-feedstock).

Some questions:
- what would be the disadvantages of using `category_encoders`, besides reduced builtin functionality?
- assuming that making something "tpot-ready" is "just" adding appropriate bounds to parameters a la `regressor.py` and `classifier.py`, could these be advertised in `entry_points`, either on `tpot` or on the upstream say:
```python
# setup.py
setup(
  entry_points=[
     "tpot.classifier": [
         "category_encoders": "tpot.config.extras.category_encoders:classifiers"
     ]
  ]

# tpot/config/extras/category_encoders.py
def classifiers():
    return {
        'category_encoders.one_hot.OneHotEncoder': {
          ...
        }
 },
```
  - could the metadata about the config flavors (sparse, lite, etc) be captured inside the config entry vs. as whole separate sets 
  - Aside from a runtime dependency on `setuptools` (or the lighter-weight `entrypoints`), what are the disadvantages of adopting `entry_points` for discovery?
- Are there any heuristics for whether a classifier/regressor is "interesting" to tpot? would any of the other [scikit-learn-contrib](https://github.com/scikit-learn-contrib) things match those criteria?

Looking forward to digging into this more!

rhiever:
Hi @bollwyvl, thanks for your interest in TPOT and this PR. This PR has been stagnant for a while because it entailed some bigger changes to TPOT than we originally suspected. Before this PR can really be merged, we need to make sure that it *at least* doesn't make TPOT perform worse, which entails a sizable benchmark study.

The core idea of this PR is to encourage TPOT to treat categorical and continuous features differently, as TPOT has a tendency to apply preprocessing operations wholesale to the entire dataset, when some preprocessing operations weren't designed with such an application in mind. As such, this PR entails a deeper change to TPOT than adding new preprocessing operators: we need to devise a mechanism for TPOT to optionally *only* pull continuous or categorical features from a dataset without causing significant performance degradation.

I think there are many integrations of TPOT and scikit-learn-contrib packages that could be interesting. We prepackage a few different flavors of TPOT as you noted (normal, light, sparse, etc.) but ultimately wanted to leave the heavy customization of TPOT to users via the [custom config interface](http://rhiever.github.io/tpot/using/#customizing-tpots-operators-and-parameters) in TPOT. We would be interested in integrating more operators into the prepackaged TPOT configurations if they show consistent promise for improving TPOT's performance on a variety of problems. That would again entail a sizable benchmark study, either within or outside of TPOT.

bollwyvl:
@rhiever thanks for the thorough guidance and patience!

To the selection end, indeed: that sounds like a tough problem. One thing I liked about, say, the `category_encoders` approach is that explicitly allows for passing in columns to encode, though as suggested #602, arbitrary callables aren't supported... having a callable that accepts the data itself is getting pretty murky indeed...

Again, as I am starting out and my goal is "learn about the landscape", my reaction in terms of possible pipeline members is "more is better." Of course, if the goal is "generate a good pipeline before the end of the universe," every additional possible pipeline member expands the search space even more. 

To the learning end, perhaps an annotated, discovery-focused "kitchen sink" could be useful, which encouraged the contribution of new configuration members.

Indeed, melting my brain further, one could imagine said benchmark study using such a configuration, along with [metafeatures](https://github.com/rhiever/sklearn-benchmarks), etc. to burn _even more_ coal and identify correlation between datasets and approaches.

rhiever:
>one could imagine said benchmark study using such a configuration, along with metafeatures, etc. to burn even more coal and identify correlation between datasets and approaches.

Exactly! This is the main idea behind meta-learning. I made an early attempt to do this in TPOT (paper [here](https://arxiv.org/abs/1607.08878)), but that approach didn't work well. I would encourage all work toward that goal.

mattspinelli:
What is the status on this PR considering it is now 6 months old? Is something holding this up from being merged? I could benefit from this as I have a dataset with mixed categorical basically continuous data.

weixuanfu:
OK, I will fix the conflicts soon and then merge it into dev branch

mattspinelli:
@weixuanfu 

Reading this PR closer it seems there was hesitation to include it because a benchmark study hasn't been done to access if this helps improve performance overall and talk. If there was a flag to to make the categorical/continuous selector optional (default not enabled) then this probably would be the least invasive way to incorporate this feature. I don't know if this optional flag would be the "deeper change" @rhiever envisioned.

weixuanfu:
For now, I think a better way is just to use confg_dict parameter to include them. I will provide a demo for it.

weixuanfu:
@mattspinelli 

I will merge the branch to TPOT dev branch soon. You could install the tweaked version of TPOT via the commend below:

```
pip install --upgrade --no-deps --force-reinstall git+https://github.com/EpistasisLab/tpot.git@development
```

Below is a demo for using these two new transformer:

```
from tpot.config import classifier_config_dict
import numpy as np
# add 2 operators into the classifier_config_dict
classifier_config_dict['tpot.builtins.CategoricalSelector'] = {
    'threshold': [10],
    'minimum_fraction': [0.05, 0.1, 0.15, 0.2, 0.25],
    'sparse': [False]
}

classifier_config_dict['tpot.builtins.ContinuousSelector'] = {
    'threshold': [10],
    'svd_solver': ['randomized'],
    'iterated_power': range(1, 11)
}

# load data
iris = load_iris()
    iris.target.astype(np.float64), train_size=0.75, test_size=0.25)

tpot = TPOTClassifier(generations=5, population_size=50, verbosity=2, config_dict=classifier_config_dict)
tpot.fit(X_train, y_train)
```

[More info for using `config_dict`](https://epistasislab.github.io/tpot/using/#customizing-tpots-operators-and-parameters)

Issue #585, opened 2017-10-05
Keeping track of more statistics of individuals


bartdp1:
TPOT recently got my interest, and i must say i really like it, both from conceptual as well as performance point of view.  To do some in depth analyses on the inner workings and performance of the algorithm, it would be great to have some more statistics on the individuals that TPOT examines along the way. These could for example be added to the '_evaluated_individuals_' attribute, where currently the number of components and the internal cross validation score is available.
For example, interesting things to have available would be:
- Generation of occurrence
- Cumulative number of crossover operations applied to the individual (meaning the number of operations applied to the individual and its predecessors)
- Cumulative number of mutation operations applied to the individual, maybe split per type of mutation
- (List of) String representation(s) of predecessor(s)
And maybe more.
 
For all of the mentioned statistics i think it would be sufficient to know them at the moment a pipeline is encountered for the first time, ignoring when a pipeline might be found again in a later generation.

Although i am not extremely skilled in Python, i want to make an effort (and try) to implement this myself, so i dove into the structure of TPOT and DEAP a bit. I was thinking about maintaining and updating the statistics in the individual objects themselves as attributes, updating them with crossover, mutation and precedence operations, and write them to the '_evaluated_individuals_' dict upon evaluation of the individuals. This would result in the '_evaluated_individuals_' begin similar to what it is currently, just with an expanded statistics tuple.

I'm curious about what you think of the idea, and about the proposed implementation.

rhiever:
Thank you for your interest in TPOT, @bartdp1. This is an interesting feature that could be useful to TPOT power users and researchers. My fear for this feature is that it would have considerable overhead for the non-research version of TPOT, both in terms of memory/computation and in terms of code integration.

For example, to track the list of predecessors, we would need to store an additional object in TPOT that tracks the ancestry of every individual, which can grow quite large. Then to store the ancestry for one individual, we would need to traverse down that ancestry tree and store a list of strings for every evaluated individual.

If this feature is proposed for a research version of TPOT for a research project, I don't think I would have these reservations.

Perhaps @weixuanfu can add some thoughts about the feature?

bartdp1:
[...]

PR #587, opened 2017-10-06
Implemented statistics tracking

bartdp1:
## What does this PR do?

self.evaluated_individuals is now a dict with key being the string representation of the individual and the value being a dict containing statistics about the pipeline at the time it was evaluated.

The DEAP individuals have a new attribute `statistics`, which is a dict containing current information about the individual. Initialization is done with a new function `initialize_stats_dict`, which is called in the `eaMuPlusLambda` function. 

Statistics are updates as individuals advance to the next generation, are mutated or crossed. More specifically

- generation - this is updated in the `eaMuPlusLambda` function as the iteration for a new generation is started.
- crossover_count - this is updated in the `self._mate_operator` function.
- mutation_count - this is updated in the `self._random_mutation_operator` function.
- predecessor - this is updated both the `self._mate_operator` and  `self._random_mutation_operator` 

a new function in the base class is introduced `self._final_stats` which consolidates the mentioned statistics with the operator count and cross validation score, and is called at any occasion where `self.evaluated_individuals` is updated. 

The `self_evaluate_individuals` function had some minor changes in order to return the right statistics.

`self._preprocess_individuals` returns an extra dict, where key is the string representation of the individuals, and the value is the dict containing the statistics. `_update_evaluated_individuals_` has this dict added to it's arguments. 

furthermore, the tests in tpot_tests.py are adjusted to account for the changes. stats_test.py is added to test the specific functionality regarding the stats.

Follow the changes as mentioned above.

Tests for the added functionality can be found in `stats_test.py.` Suggestions for tests are welcome.

#585

## Questions:

Description of the evaluated_individuals_ should be adjusted.
No.

### DIFF
--- a/tests/stats_test.py
+++ b/tests/stats_test.py
@@ -1,0 +1,104 @@
+from tpot import TPOTClassifier
+from tpot.gp_deap import initialize_stats_dict
+
+from deap import creator
+
+import random
+
+
+def test_dict_initialization():
+    """Asserts that gp_deap.initialize_stats_dict initializes individual statistics correctly"""
+    tpot = TPOTClassifier()
+    tb = tpot._toolbox
+
+    test_ind = tb.individual()
+    initialize_stats_dict(test_ind)
+
+    assert test_ind.statistics['generation'] == 0
+    assert test_ind.statistics['crossover_count'] == 0
+    assert test_ind.statistics['mutation_count'] == 0
+    assert test_ind.statistics['predecessor'] == ('ROOT',)
+
+
+def test_mate_operator_stats_update():
+    """Assert that self._mate_operator updates stats as expected."""
+    tpot_obj = TPOTClassifier()
+    ind1 = creator.Individual.from_string(
+        'KNeighborsClassifier('
+        'BernoulliNB(input_matrix, BernoulliNB__alpha=10.0, BernoulliNB__fit_prior=False),'
+        'KNeighborsClassifier__n_neighbors=10, '
+        'KNeighborsClassifier__p=1, '
+        'KNeighborsClassifier__weights=uniform'
+        ')',
+        tpot_obj._pset
+    )
+    ind2 = creator.Individual.from_string(
+        'KNeighborsClassifier('
+        'BernoulliNB(input_matrix, BernoulliNB__alpha=10.0, BernoulliNB__fit_prior=True),'
+        'KNeighborsClassifier__n_neighbors=10, '
+        'KNeighborsClassifier__p=2, '
+        'KNeighborsClassifier__weights=uniform'
+        ')',
+        tpot_obj._pset
+    )
+
+    initialize_stats_dict(ind1)
+    initialize_stats_dict(ind2)
+
+    # Randomly mutate the statistics
+    ind1.statistics["crossover_count"] = random.randint(0, 10)
+    ind1.statistics["mutation_count"] = random.randint(0, 10)
+    ind2.statistics["crossover_count"] = random.randint(0, 10)
+    ind2.statistics["mutation_count"] = random.randint(0, 10)
+
+    # set as evaluated pipelines in tpot_obj.evaluated_individuals_
+    tpot_obj.evaluated_individuals_[str(ind1)] = tpot_obj._final_stats(2, 0.99, ind1.statistics)
+    tpot_obj.evaluated_individuals_[str(ind2)] = tpot_obj._final_stats(2, 0.99, ind2.statistics)
+
+    # Doing 10 tests
+    for _ in range(10):
+        offspring1, offspring2 = tpot_obj._mate_operator(ind1, ind2)
+
+        assert offspring1.statistics['crossover_count'] == ind1.statistics['crossover_count'] + ind2.statistics[
+            'crossover_count'] + 1
+        assert offspring1.statistics['mutation_count'] == ind1.statistics['mutation_count'] + ind2.statistics[
+            'mutation_count']
+        assert offspring1.statistics['predecessor'] == (str(ind1), str(ind2))
+
+        # Offspring replaces on of the two predecessors
+        # Don't need to worry about cloning
+        if random.random() < 0.5:
+            ind1 = offspring1
+        else:
+            ind2 = offspring1
+
+
+def test_mut_operator_stats_update():
+    """Asserts that self._random_mutation_operator updates stats as expected."""
+    tpot_obj = TPOTClassifier()
+    ind = creator.Individual.from_string(
+        'KNeighborsClassifier('
+        'BernoulliNB(input_matrix, BernoulliNB__alpha=10.0, BernoulliNB__fit_prior=False),'
+        'KNeighborsClassifier__n_neighbors=10, '
+        'KNeighborsClassifier__p=1, '
+        'KNeighborsClassifier__weights=uniform'
+        ')',
+        tpot_obj._pset
+    )
+
+    initialize_stats_dict(ind)
+
+    ind.statistics["crossover_count"] = random.randint(0, 10)
+    ind.statistics["mutation_count"] = random.randint(0, 10)
+
+    # set as evaluated pipelines in tpot_obj.evaluated_individuals_
+    tpot_obj.evaluated_individuals_[str(ind)] = tpot_obj._final_stats(2, 0.99, ind.statistics)
+
+    for _ in range(10):
+        offspring, = tpot_obj._random_mutation_operator(ind)
+
+        assert offspring.statistics['crossover_count'] == ind.statistics['crossover_count']
+        assert offspring.statistics['mutation_count'] == ind.statistics['mutation_count'] + 1
+        assert offspring.statistics['predecessor'] == (str(ind),)
+
+        ind = offspring
--- a/tests/stats_test.py
+++ b/tests/stats_test.py
@@ -52,8 +52,8 @@ def test_mate_operator_stats_update():
     ind2.statistics["mutation_count"] = random.randint(0, 10)
 
     # set as evaluated pipelines in tpot_obj.evaluated_individuals_
-    tpot_obj.evaluated_individuals_[str(ind1)] = tpot_obj._final_stats(2, 0.99, ind1.statistics)
-    tpot_obj.evaluated_individuals_[str(ind2)] = tpot_obj._final_stats(2, 0.99, ind2.statistics)
+    tpot_obj.evaluated_individuals_[str(ind1)] = tpot_obj._combine_individual_stats(2, 0.99, ind1.statistics)
+    tpot_obj.evaluated_individuals_[str(ind2)] = tpot_obj._combine_individual_stats(2, 0.99, ind2.statistics)
 
     # Doing 10 tests
     for _ in range(10):
@@ -90,7 +90,7 @@ def test_mut_operator_stats_update():
     ind.statistics["mutation_count"] = random.randint(0, 10)
 
     # set as evaluated pipelines in tpot_obj.evaluated_individuals_
-    tpot_obj.evaluated_individuals_[str(ind)] = tpot_obj._final_stats(2, 0.99, ind.statistics)
+    tpot_obj.evaluated_individuals_[str(ind)] = tpot_obj._combine_individual_stats(2, 0.99, ind.statistics)
 
     for _ in range(10):
         offspring, = tpot_obj._random_mutation_operator(ind)
--- a/tests/stats_test.py
+++ b/tests/stats_test.py
@@ -59,10 +59,8 @@ def test_mate_operator_stats_update():
     for _ in range(10):
         offspring1, offspring2 = tpot_obj._mate_operator(ind1, ind2)
 
-        assert offspring1.statistics['crossover_count'] == ind1.statistics['crossover_count'] + ind2.statistics[
-            'crossover_count'] + 1
-        assert offspring1.statistics['mutation_count'] == ind1.statistics['mutation_count'] + ind2.statistics[
-            'mutation_count']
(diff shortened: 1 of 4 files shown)


rhiever:
I see. This is a pretty smart way to track lineage and related stats about each individual. I suppose since we don't really allow the same pipeline to be generated via crossover/mutation once it's been previously evaluated (barring the situation where a crossover/mutation can't generate an entirely new individual within 10 attempts), there's really no need to maintain pointers from every offspring to its parents.

Well done and thank you for the PR, @bartdp1. I'll allow @weixuanfu some time to review this PR as well. I've added my comments.

tpot/base.py
```diff
@@ -1059,9 +1066,14 @@ def _evaluate_individuals(self, individuals, features, target, sample_weight=Non
                 for val in tmp_result_scores:
                     result_score_list = self._update_val(val, result_score_list)
 
-        self._update_evaluated_individuals_(result_score_list, eval_individuals_str, operator_counts)
+        self._update_evaluated_individuals_(result_score_list, eval_individuals_str, operator_counts, stats_dicts)
+
+        def get_optimization_values(individual):
```
rhiever: `get_optimization_values` should be an internal function to TPOT, no? Otherwise it's constantly being re-defined every time `_evaluate_individuals` is being called.

rhiever: Actually, is this function needed at all? It's only used once, below.

tpot/gp_deap.py
```diff
@@ -218,6 +233,10 @@ def eaMuPlusLambda(population, toolbox, mu, lambda_, cxpb, mutpb, ngen, pbar,
         if per_generation_function is not None:
             per_generation_function()
 
+        # Update generation statistic for all individuals
+        for ind in population:
+            ind.statistics['generation'] = gen
```
rhiever: This update procedure might be problematic for individuals that stick around from the previous generation. In our current GP algorithm, we don't have complete turnover between each iteration of the GP.

Need to add a check that each individuals' `generation` statistic hasn't already been set to a non-default value.

tpot/gp_deap.py
```diff
@@ -199,6 +210,10 @@ def eaMuPlusLambda(population, toolbox, mu, lambda_, cxpb, mutpb, ngen, pbar,
     logbook = tools.Logbook()
     logbook.header = ['gen', 'nevals'] + (stats.fields if stats else [])
 
+    # Initialize statistics dict
```
rhiever: More descriptive comment, please.

tpot/gp_deap.py
```diff
@@ -149,6 +149,17 @@ def varOr(population, toolbox, lambda_, cxpb, mutpb):
 
     return offspring
 
+def initialize_stats_dict(individual):
+    '''
+    Initializes the stats dict for individual
+    :param individual:
+    :return:
```
rhiever: Please make sure the function docstrings match the docstring format that we use in the rest of TPOT.

tpot/base.py
```diff
@@ -1230,6 +1257,10 @@ def _random_mutation_operator(self, individual, allow_shrink=True):
             ind = self._toolbox.clone(individual)
             offspring, = mutator(ind)
             if str(offspring) not in self.evaluated_individuals_:
+                # Update statistics
```
rhiever, line 1322: More descriptive comment, please.

tpot/base.py
```diff
@@ -1184,8 +1203,16 @@ def _mate_operator(self, ind1, ind2):
...
             ind1_copy, ind2_copy = self._toolbox.clone(ind1),self._toolbox.clone(ind2)
             offspring, offspring2 = cxOnePoint(ind1_copy, ind2_copy)
+
             if str(offspring) not in self.evaluated_individuals_:
                 # We only use the first offspring, so we do not care to check uniqueness of the second.
+
+                # update statistics
+                offspring.statistics['predecessor'] = (str(ind1), str(ind2))
+                offspring.statistics['mutation_count'] = ind1.statistics['mutation_count'] + ind2.statistics[
+                    'mutation_count']
```
rhiever: No need for line break here; we use a lax version of the PEP8 guidelines.

tpot/base.py
```diff
@@ -1184,8 +1203,16 @@ def _mate_operator(self, ind1, ind2):
...
+                offspring.statistics['crossover_count'] = ind1.statistics['crossover_count'] + ind2.statistics[
+                    'crossover_count'] + 1
```
rhiever: No need for line break here; we use a lax version of the PEP8 guidelines.

tpot/base.py
```diff
@@ -1001,6 +1001,13 @@ def _stop_by_max_time_mins(self):
             if total_mins_elapsed >= self.max_time_mins:
                 raise KeyboardInterrupt('{} minutes have elapsed. TPOT will close down.'.format(total_mins_elapsed))
 
+    def _final_stats(self, operator_count, cv_score, individual_stats):
```
rhiever: Maybe a more descriptive function name: `_combine_individual_stats`?

tpot/base.py
```diff
@@ -1001,6 +1001,13 @@ def _stop_by_max_time_mins(self):
...
+        """Combine the stats with operator count and cv score and preprare to be written to _evaluated_individuals"""
```
rhiever: Please annotate the parameters and output(s) of this function in the docstrings.

tests/stats_test.py
```diff
@@ -0,0 +1,104 @@
+from tpot import TPOTClassifier
```
rhiever, line 1: Thank you for the unit tests. 👍

tests/stats_test.py
```diff
@@ -0,0 +1,104 @@
...
+    # set as evaluated pipelines in tpot_obj.evaluated_individuals_
+    tpot_obj.evaluated_individuals_[str(ind1)] = tpot_obj._final_stats(2, 0.99, ind1.statistics)
+    tpot_obj.evaluated_individuals_[str(ind2)] = tpot_obj._final_stats(2, 0.99, ind2.statistics)
+
+    # Doing 10 tests
+    for _ in range(10):
+        offspring1, offspring2 = tpot_obj._mate_operator(ind1, ind2)
+
+        assert offspring1.statistics['crossover_count'] == ind1.statistics['crossover_count'] + ind2.statistics[
+            'crossover_count'] + 1
```
rhiever: No need for this newline.

tests/stats_test.py
```diff
@@ -0,0 +1,104 @@
...
+        assert offspring1.statistics['mutation_count'] == ind1.statistics['mutation_count'] + ind2.statistics[
+            'mutation_count']
```
rhiever: No need for this newline.

tpot/gp_deap.py
weixuanfu: What is this `ind.statistics['generation']` use for? I think some description about it is needed. If it is about which generation(s) has this individual, it should be a list instead.

tpot/gp_deap.py
weixuanfu: Please add more detailed docstring here too.

tpot/base.py
```diff
@@ -1103,7 +1116,7 @@ def _preprocess_individuals(self, individuals):
             individual_str = str(individual)
             sklearn_pipeline_str = generate_pipeline_code(expr_to_tree(individual, self._pset), self.operators)
             if sklearn_pipeline_str.count('PolynomialFeatures') > 1:
-                self.evaluated_individuals_[individual_str] = (5000., -float('inf'))
+                self.evaluated_individuals_[individual_str] = self._final_stats(5000., -float('inf'), individual.statistics)
```
weixuanfu: Please reformat this line as the line 1144.

tpot/gp_deap.py
rhiever: I believe the `generation` statistic is meant to track the generation that the individual first appeared.

Good point though, @weixuanfu. Since we're moving forward with this change, this PR should add some documentation (in the docstrings and the API documentation) about these new statistics.

tpot/base.py
```diff
@@ -1125,17 +1138,21 @@ def _preprocess_individuals(self, individuals):
...
                     continue
                 eval_individuals_str.append(individual_str)
                 sklearn_pipeline_list.append(sklearn_pipeline)
 
-        return operator_counts, eval_individuals_str, sklearn_pipeline_list
+        return operator_counts, eval_individuals_str, sklearn_pipeline_list , stats_dicts
 
 
-    def _update_evaluated_individuals_(self, result_score_list, eval_individuals_str, operator_counts):
+    def _update_evaluated_individuals_(self, result_score_list, eval_individuals_str, operator_counts, stats_dicts):
```
weixuanfu: Please add more detailed docstring here.

tpot/base.py
bartdp1: The function is indeed only used once below, but i find the list compresension in which the function is called better readable in this way. Will change it

tpot/gp_deap.py
bartdp1: Since the information in statistics only gets written to `evaluted_individuals_` upon evaluation, and an individual is evaluated in the generation that it first occurs, the statistics reported in `evaluated_individuals_` is always the information of the individual at the time of evaluation. In this light i did not build in any checks here, however, for debugging purposes it might be a good idea to indeed update `generation` only when it concerns a new individual, and not update for individuals that traverse from the previous generation. 

I can implement this if you like, let me know.

tpot/gp_deap.py
bartdp1: Will do. Is this any established format, then i can set my IDE settings which will save me some time.

tpot/base.py
bartdp1: Ok. Think this will just be flipping a switch in the IDE.

tpot/gp_deap.py
bartdp1: Moreover, there is some discrepancy in the format used. E.g. comparing `mutate_random_individual` and `varOr` in `gp_deap`.

tpot/base.py
```diff
@@ -630,11 +625,11 @@ def _update_top_pipeline(self):
...
-                                  'process. This could be because the data was '
-                                  'not formatted properly, or because data for '
-                                  'a regression problem was provided to the '
-                                  'TPOTClassifier object. Please make sure you '
-                                  'passed the data to TPOT correctly.')
+                                   'process. This could be because the data was '
+                                   'not formatted properly, or because data for '
+                                   'a regression problem was provided to the '
+                                   'TPOTClassifier object. Please make sure you '
+                                   'passed the data to TPOT correctly.')
```
rhiever, line 687: Thank you for the various formatting changes. However, as a note for future PRs, beware of making numerous small changes like these ones when they are unrelated to the core PR. Numerous small, unrelated changes can cause merge conflicts if another PR happens to be changing those lines as well. I think we can make it work this time.

tpot/gp_deap.py
rhiever: Because the behavior you described is a function of other parts of TPOT that are not directly related to the statistics tracking, and that behavior could change at any point and break this assumption, let's add an explicit check here to make sure that persistent individuals don't have their `generation` statistic updated. Please also add a comment above that check describing why we're doing that, so future developers aren't confused by the logic.

tpot/base.py
rhiever: Ping on this small change request, @bartdp1.

tpot/gp_deap.py
bartdp1: @rhiever Got it. How about incrementing the generation statistic along with the other statistics, at the moment the new individual is generated by mutation/cross-over? @rhiever. Am i missing any cases this way?

tpot/base.py
bartdp1: Missed it, Will do tomorrow

tpot/base.py
bartdp1: Hmm, sorry for this. Think i might have accidently hit ‘auto format’ in my ide. Hopefully wont be harmfull in any way. Let me know if i should roll back Some things.

rhiever: No worries. I'm pretty sure we can sort out any merge conflicts this time because there's only one other PR being actively developed (just merged into the dev branch).

tpot/base.py
bartdp1: It's already there, in one of the last commits :)

tpot/gp_deap.py
rhiever: Yes, we could do that as long as the `generation` is available to that function.

rhiever: Yes, we could do that as long as the `generation` is available to that function.

bartdp1: Ah i see, it's not Always generation = generation + 1. The way i've implemented the check:
Mutation and crossover 'invalidate' the generation stat. After varOr is run, the generation stat is updaten when invalid.

bartdp1:
@rhiever @weixuanfu. Seems like the PR times out on some unit test. Found issue #588, which adresses the same problem. How can i best proceed with integrating the fix? Copying it and comitting it to the PR? Thanks in advance

weixuanfu:
@bartdp1 you may try the git commands below for updating these expensive unit tests to fix this issue.

```
git cherry-pick 68e42f2136ec43200964d355c5c68288af85e7f5
git cherry-pick 9b428d73e5724ecb61bebc40f69b8628ecc2474a
```

weixuanfu:
You may need [this commit](https://github.com/rhiever/tpot/pull/594/commits/9b428d73e5724ecb61bebc40f69b8628ecc2474a)

bartdp1:
@rhiever @weixuanfu. I think i got all your feedback integrated now, please let me know if there are some more modifications i can make :).

rhiever:
Merged. Thanks again for your great PR, @bartdp1!

Issue #588, opened 2017-10-06
Unit tests freezing and timing out on Python 2.7 / Windows


rhiever:
The TPOT unit tests have been timing out and failing for Python 2.7 / Windows. Both on the `master` and `development` branch, CI gets stuck on the following unit test:

`Assert that varOr() applys crossover only and removes CV scores in offsprings.`

Eventually the AppVeyor times out the process and the unit tests fail.

It looks like [this](https://github.com/rhiever/tpot/commit/8b2b27b771aea1a60ebe2abc1decedd5f08f7bb3) was the breaking commit, though it's unclear why a small change to the docs caused this behavior all of the sudden.

weixuanfu:
Failed for Python 2.7 in Linux too but freezing in a different unit test. (check [this](https://travis-ci.org/rhiever/tpot/jobs/284386435)). I will look into this.

weixuanfu:
It also happened in Python 3.6 in the same unit test (check [this](https://travis-ci.org/rhiever/tpot/builds/284397636?utm_source=github_status&utm_medium=notification)).

I think there is a infinity loop in [_random_mutation_operator](https://github.com/rhiever/tpot/blob/master/tpot/base.py#L1195). The issue is about [L1239-L1241](https://github.com/rhiever/tpot/blob/master/tpot/base.py#L1239-L1241)

weixuanfu:
I pushed [a commit](https://github.com/rhiever/tpot/pull/580/commits/7a182e297ae5caa40e7cbafe3e2ad9bb31da5346) in #580 for fixing this infinity loop.

I think in rare cases the mutation operator cannot generated a valid and new individual and then it goes into a infinity loop.

weixuanfu:
I still did not figure out why this freezing happened in CI. I cannot reproduce it in my environments with python 2.7 in macOS, Linux and Windows. It is very odd.

PR #590, opened 2017-10-09
Check unit tests in python 2.7

weixuanfu:
## What does this PR do?

Since I cannot reproduce the issue #588 in my environments with python 2.7 in macOS, Linux and Windows, I suspect that maybe unit tests with running `cross_eval_score` on a expensive pipeline, (e.g. two `PolynomialFeatures` in a pipeline). I submit this PR for fixing this issue and also test in CI environment.


rhiever:
It works! Huzzah!

weixuanfu:
Great, we have much more unit tests in version 0.9 than 0.8. Maybe they need more resources than before and also I think we need refine some unit tests and make them not use too much resources.

Issue #605, opened 2017-10-25
pywin32 can pip install only up to 3.3


mikister:
While following the installation docs I tried using
```shell
pip install pywin32
```
but got an error saying that the required version couldn't be found.
On [PyPi](https://pypi.python.org/pypi/pywin32) it says it is available up to Python version 3.3.

 I use:
1. Windows 10
2. python 3.6.3
3. install with pip

After some searching on Stack Overflow, there was a suggestion to install it from [Unofficial Windows Binaries for Python Extension Packages](http://www.lfd.uci.edu/~gohlke/pythonlibs/).

And so I did:

1. Install from that page, whilst following their instructions
2. Needs restarting the computer to work
3. Ran TPOTClassifier example from Readme.md and it worked fine

So I thought that this should be mentioned in the docs.

weixuanfu:
Could you please try `conda install -c anaconda pywin32`?

mikister:
Here's what I got @weixuanfu :
![screenshot_2](https://user-images.githubusercontent.com/16947397/32046135-6851d882-ba42-11e7-9ed3-a55b6148e27a.png)

weixuanfu:
Thanks, I think windows users may need use conda to install it instead of pypi.

mikister:
@weixuanfu 
You should be able to install it from pypi with python 2.x but not with python 3.x, here you must use conda, or you can install pywin32 from a wheel.

weixuanfu:
@mikister Agree. I just updated the PR. Thank you for catching this issue and suggestion.

weixuanfu:
I close this issue since the PR was merged to dev branch. Please feel free to re-open the issue (or comment further) if you have any more questions

PR #606, opened 2017-10-26
Update installation guide of pywin32

weixuanfu:
Just update installation guide for pywin32 and require windows users to install pywin32 from conda instead of pypi.

Related issue #605


weixuanfu:
Not sure why AppVeyor build failed. @rhiever

Issue #616, opened 2017-11-06
Failure to periodically save pipeline if the directory doesn't exist


mlapierre:
If the periodic_checkpoint_folder doesn't exist an error is thrown

## Context of the issue

In the spirit of automation, it would be nice if the periodic_checkpoint_folder were created for us if it doesn't exist. At the moment an error is thrown and no checkpoints are saved.

## Process to reproduce the issue

Create a TPOT instance with a nonexistent folder as the periodic_checkpoint_folder

## Expected result

The folder is automatically created

## Current result

An error occurs:

> Failed saving periodic pipeline, exception:
> [Errno 2] No such file or directory: './checkpoints\\pipeline_2017.11.06_06-05-10.py'

## Possible fix

Before:
https://github.com/rhiever/tpot/blob/a303f1e9b209a7d0625c9e6017169a7e3c9bf4d1/tpot/base.py#L864

Include something like:
```python
    try:
        os.makedirs(self.periodic_checkpoint_folder)
    except OSError as e:
        if e.errno == errno.EEXIST and os.path.isdir(self.periodic_checkpoint_folder):
            pass
```

If this would be useful I'm happy to submit a PR.

rhiever:
@mlapierre, we'd be happy to review and merge a PR for this enhancement to the periodic checkpoint folder feature.

weixuanfu:
@mlapierre I closed this PR since the PR already merged to dev branch! Thank you!

PR #617, opened 2017-11-06
Create the periodic checkpoint folder if it doesn't already exist

mlapierre:
Create the periodic_checkpoint_folder first (and all the parent folders) before trying to write to it.

Addresses #616

### DIFF
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -34,7 +34,6 @@ from datetime import datetime
 from multiprocessing import cpu_count
 import os
 import re
-import errno
 from tempfile import mkdtemp
 from shutil import rmtree
 
@@ -398,11 +397,11 @@ class TPOTBase(BaseEstimator):
                     self.config_dict = config.tpot_config
                 else:
                     raise ValueError(
-                                    'Could not find "tpot_config" in configuration file {}. '
-                                    'When using a custom config file for customizing operators '
-                                    'dictionary, the file must have a python dictionary with '
-                                    'the standardized name of "tpot_config"'.format(config_dict)
-                                    )
+                        'Could not find "tpot_config" in configuration file {}. '
+                        'When using a custom config file for customizing operators '
+                        'dictionary, the file must have a python dictionary with '
+                        'the standardized name of "tpot_config"'.format(config_dict)
+                    )
         else:
             self.config_dict = self.default_config_dict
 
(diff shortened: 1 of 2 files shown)


tpot/base.py
```diff
@@ -929,6 +932,15 @@ def _save_periodic_pipeline(self):
...
             self._update_pbar(pbar_num=0, pbar_msg='Failed saving periodic pipeline, exception:\n{}'.format(str(e)[:250]))
 
+    def _create_periodic_checkpoint_folder(self):
+        try:
+            os.makedirs(self.periodic_checkpoint_folder)
+        except OSError as e:
+            if e.errno == errno.EEXIST and os.path.isdir(self.periodic_checkpoint_folder):
+                pass
+            else:
+                raise
```
weixuanfu: I think  a ValueError with a message about this issue need be raised here

tpot/base.py
```diff
@@ -929,6 +932,15 @@ def _save_periodic_pipeline(self):
         except Exception as e:
             self._update_pbar(pbar_num=0, pbar_msg='Failed saving periodic pipeline, exception:\n{}'.format(str(e)[:250]))
 
+    def _create_periodic_checkpoint_folder(self):
+        try:
+            os.makedirs(self.periodic_checkpoint_folder)
+        except OSError as e:
+            if e.errno == errno.EEXIST and os.path.isdir(self.periodic_checkpoint_folder):
+                pass
```
weixuanfu: We need add a warning message to let user know if the folder exists or tpot make a new folder.

weixuanfu (commented):
Thank you for this PR.

tpot/base.py
mlapierre: This will now looks something like:

> Failed saving periodic pipeline, exception:
> Failed creating the periodic_checkpoint_folder:
> [WinError 5] Access is denied: './tmp'

As the error bubbles up. Is that ok?

tpot/base.py
mlapierre: I added a line to output a message via the progress bar when the folder is created. I didn't replace the `pass` with a message because I'd expect that to be the default path - the user will probably create a folder so I'm not sure they need a message then, and if they don't create one we probably only want to show the message about the automatic folder creation once. Is that ok?

weixuanfu: oic, OK, just a friendly message when a new folder is created would be fine!

tpot/base.py
weixuanfu: looks good! Thanks.

rhiever:
Looks good. Thanks for this fix, @mlapierre!

Issue #620, opened 2017-11-08
Fixed parameter


korcky:
## Context of the issue

Sometimes I need to fixed some of model's parameters (without any changing in model's code) and when this parameter is value -- it's work fine (most of the time used non default value). Like this:

```python
config_dict = {
    'tpot_test.Test': {
        'a': np.arange(-1, 1, 0.1),
        'b': [1]
    }
}
```

But when this parameter should be method or class TPOT always used DEFAULT value

## Code example

```python
import numpy as np
from sklearn.base import RegressorMixin
from sklearn.metrics import r2_score
from tpot import TPOTRegressor


class Test(RegressorMixin):
    def __init__(self, a=None, b=None):
        print(a, b)
        self.a = a
        self.b = b

    def fit(self, X, y):
        return

    def predict(self, X):
        if self.b is r2_score:
            return self.a * X + 1
        else:
            return self.a * X

    def score(self, X, y, sample_weight=None):
        return r2_score(y_true=y, y_pred=self.predict(X))

    def get_params(self, deep=None):
        return {'a': self.a, 'b': self.b}


config_dict = {
    'tpot_test.Test': {
        'a': np.arange(-1, 1, 0.1),
        'b': [r2_score]
    }
}

tpot = TPOTRegressor(generations=5, population_size=5, cv=5, verbosity=1,
                     config_dict=config_dict)

tpot.fit(np.arange(0, 10, 1).reshape(-1, 1), (np.arange(0, 5, 0.5) + 1).reshape(-1, 1))
```

Result of `print(a, b)` always: `<value> None` (or sometimes `None None`)

weixuanfu:
Is this the latest version of TPOT (v0.9)? I think I removed the DEFAULT in this [commit](https://github.com/rhiever/tpot/commit/5d6b73184c02600a85476856b9e58be6fd85d5de).

korcky:
Oh, yeah, I forgot to update TPOT. It's fix problem with value parameter (now DEFAULT disappeared), but when using method or class exception appear:

## Code

```python
import numpy as np
from sklearn.base import RegressorMixin
from sklearn.metrics import r2_score
from tpot import TPOTRegressor


class Test(RegressorMixin):
    def __init__(self, a=None, b=None):
        print(a, b)
        self.a = a
        self.b = b

    def fit(self, X, y):
        return

    def predict(self, X):
        if self.b is r2_score:
            return self.a * X + 1
        else:
            return self.a * X

[...]

PR #622, opened 2017-11-14
Refined operator utilities

weixuanfu:
## What does this PR do?

1. Refined operator utilities to make let TPOT to tell whether a nested parameter is a callable function or sklearn's estimators instead of checking the parameter name is `score_func`
2. Add and refine unit tests for this refined operator utilities

operator_utils.py

## Any background context you want to provide? What are the relevant issues?

#620

## Questions:

### DIFF
--- a/tests/export_tests.py
+++ b/tests/export_tests.py
@@ -38,11 +38,16 @@ from deap import creator
 
 from nose.tools import assert_raises, assert_equal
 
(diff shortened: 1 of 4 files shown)


weixuanfu:
Hmm, it seems that the lastest [deap 1.2](http://deap.readthedocs.io/en/master/) may has a new API

PR #641, opened 2017-12-18
Fix "stopit" and "pandas" required version

jonathansantilli:
Before this commit, executing  `pip install -r requirements.txt` cause an error due invalid version syntax

## What does this PR do?
Fix the requirements file syntax

Looking that the current version executing `pip install -r requirements.txt` does not work

`pip install -r requirements.txt` should work correctly and install the dependencies

If someone wants to install the TPOP dependencies, it won't work

## Questions:

No
No

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,5 +5,5 @@ scikit-learn==0.18.1
 scipy==0.19.0
 tqdm==4.11.2
 update-checker==0.16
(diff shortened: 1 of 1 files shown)


weixuanfu:
Oh,right, thank you for the fix! Merged!

jonathansantilli:
👍

PR #649, opened 2018-01-05
Re-add tests to MANIFEST.in

bollwyvl:
## What does this PR do?

(Re-)Adds the tests to `MANIFEST.in`. There was a line for `tests.py` and `tests.csv`, but those aren't currently in the repo.

`python setup.py sdist | grep tests`

Ensure that the above command includes the `tests/*.{py,csv}`

First off, thanks for all the hard work!

In packaging this for conda, it's helpful to have some idea of whether the package, as installed, actually does what it says on the tin!

Further, as this package has several optional dependencies which are tested, and all available with conda, it is useful to check whether those are on the up-and-up.

## Questions:

no

no

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,3 +1,4 @@
-include README.md LICENSE tests.py tests.csv
+include README.md LICENSE
 recursive-include images *
 recursive-include tpot *.py
+recursive-include tests *.py *.csv


weixuanfu:
Thank you! It is weird that 2 ci tests on Linux failed. Any ideas?

bollwyvl:
Eek... Looks like some Network weirdness for various package managers. I'd
try a restart of the job?

Probably possible to do some more caching...

weixuanfu:
Hmm, I am not with my laptop now. Please try to push a null comment (e.g. “clean codes”) to restart the test.

rhiever:
Build restarted.

bollwyvl:
py36 without coverage failed again...

bollwyvl:
Hooray! 💚

rhiever:
Thanks @bollwyvl! Looks like the tests were timing out. Hopefully not a regular issue, otherwise we'll have to dig into the specific tests that are timing out.

bollwyvl:
Groovy. Come `0.9.2`, we'll run the tests in the packaging step on `conda-forge`!

PR #704, opened 2018-06-10
Add a rescore parameter

bopjesvla:
If the training data or scoring context changes from generation to generation, fitness metrics can't be reused for surviving pipelines. If `rescore` is set, all individuals are evaluated every generation.

### DIFF
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -100,7 +100,7 @@ class TPOTBase(BaseEstimator):
 
     def __init__(self, generations=100, population_size=100, offspring_size=None,
                  mutation_rate=0.9, crossover_rate=0.1,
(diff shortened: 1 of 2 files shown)


bopjesvla:
Fixed a silly mistake. One thing to note is that this PR currently does not re-evaluate the pareto front after every generation.

bopjesvla:
It seems that TPOT strongly assumes that data and scoring context remain constant. I'll have to go with subclassing.

weixuanfu:
We fixed random state within pipeline to ensure that the fitness metrics is the same in every generation. Also, the training data should be the same over generations in TPOT. Saving the fitness metrics for evaluated pipeline is for reducing computation time in optimization process.

Issue #708, opened 2018-06-16
Sample_weights - question about what happens, improve documentation


joseortiz3:
After sample_weights were introduced in response to [this](https://github.com/EpistasisLab/tpot/issues/310), the documentation for sample_weights was updated with this description:

> Per-sample weights. Higher weights force TPOT to put more emphasis on those points.

This leaves some questions unanswered, and I want to make sure I understand what is happening "under the hood". 

From looking at [operator_utils.py]..........(https://github.com/EpistasisLab/tpot/blob/master/tpot/operator_utils.py),

```python
def set_sample_weight(pipeline_steps, sample_weight=None):
    """Recursively iterates through all objects in the pipeline and sets sample weight.
    Parameters
    ----------
    pipeline_steps: array-like
        List of (str, obj) tuples from a scikit-learn pipeline or related object
    sample_weight: array-like
        List of sample weight
    Returns
    -------
    sample_weight_dict:
        A dictionary of sample_weight
    """
    sample_weight_dict = {}
    if not isinstance(sample_weight, type(None)):
        for (pname, obj) in pipeline_steps:
            if inspect.getargspec(obj.fit).args.count('sample_weight'):
                step_sw = pname + '__sample_weight'
                sample_weight_dict[step_sw] = sample_weight

    if sample_weight_dict:
        return sample_weight_dict
    else:
        return None
```
 ......it seems that each element in the pipeline is checked to see if it has a `.fit()` method that accepts a sample_weight argument.

If so, I have two questions regarding "missing opportunities" to pass sample_weights.

1) it might miss passing sample_weights to `fit_transform()` (but hopefully tpot never uses `fit_transform` and instead simply calls `fit` and `transform` separately)
2) it might miss passing sample_weights to some function that could use it, although I don't know of any (and there might not be any)

Also I wanted to know if sample_weights is used in tpot's scoring function, which determines preferences between pipelines?

Finally, maybe another sentence or two in the documentation to explain briefly what is happening with sample_weights would be helpful to future people having these same questions.

Thanks for tpot! Wonderful seeing this project progress over the last two years.

weixuanfu:
Thank you for your idea about sample weight.

For 2 questions, I think all the operators in TPOT should follow the scikit-learn API which means that they should has `fit`. But I will double-check it.

rhiever:
[...]

PR #717, opened 2018-06-22
Added sample_weight checking for NaN / dimension

joseortiz3:
Previously there was no sample_weight checking. Computation would then proceed until every pipeline failed. This fixes that.

## What does this PR do?

This PR would add sample_weight checking functionality to TPOT.

The checks performed are for valid conversion to float, for NaN values, and for dimension to match `target`.

It also corrects the documentation for what `_check_dataset` returns, as a "bonus" fix.

Check the changed locations in the file.

Try passing invalid sample_weight to fit(). Such as:

- [1.0, 1.0, 'oops', 1.0, .....]
- 'this was meant to be an array'
- [1.0, 1.0, np.NaN, 1.0, .....]
- an array with a length different than `target`

Found out the hard way this check is important.

#708 
[this comment](https://github.com/EpistasisLab/tpot/issues/708#issuecomment-399337754)

## Questions:

Nope.
Nope.


tpot/base.py
```diff
@@ -1004,8 +1004,18 @@ def _check_dataset(self, features, target):
...
+        # Check sample_weight
+        if sample_weight is not None:
+            try: sample_weight = np.array(sample_weight).astype('float')
+            except ValueError as e:
+                raise ValueError('sample_weight could not be converted to float array: %s' % e)
+            if np.any(np.isnan(sample_weight)):
+                raise ValueError('sample_weight contained NaN values.')
+            try: check_consistent_length(sample_weight, target)
+            except ValueError as e:
+                raise ValueError('sample_weight dimensions did not match target: %s' % e)
```
weixuanfu, line 1019: Thank you for those checks. I will test them and add some unit tests for them next week before merging this PR.

PR #719, opened 2018-06-25
Add imputation to predict proba

adithyabsk:
## What does this PR do?
Updates predict proba to impute values to prevent it from breaking from nan inputs.

tpot/base.py
  --> def predict_proba(self, features):
         ....

Try using a test input with nans in predict_proba

Predict proba was breaking before this change.

None

N/A

## Questions:
No
No


weixuanfu:
Thank you for this PR. I added unit tests and then merged the PR.

PR #724, opened 2018-07-10
Detect whether TPOT is running in a notebook

weixuanfu:
Related issue #723. Credits to @rhiever

Issue #723, opened 2018-07-09
Improve TPOT progress bar: detect whether TPOT is running in a notebook


rhiever:
This is a user experience quality issue.

TPOT should detect whether it's running in a notebook and use `tqdm_notebook` for the progress bar instead of the regular `tqdm`. This can be accomplished as simply as detecting whether TPOT is being run in a notebook during import time, and changing the `tqdm` import from `from tqdm import tqdm` to `from tqdm import tqdm_notebook as tqdm`.

See [this page](https://stackoverflow.com/questions/47211324/check-if-module-is-running-in-jupyter-or-not) and related Q&As on how to accomplish this.

PR #768, opened 2018-09-18
Use tqdm.autonotebook to automatically choose between console or notebook

weixuanfu:
Since `tqdm` v4.25.0, there is a new submodule named  `autonotebook` for automatically choosing between console or notebook versions. So I updated the version of dependency and refined the `tqdm` importing.

#723 #724


rhiever:
Nice find!

PR #811, opened 2018-11-28
Add GAMA into docs_sources (for next version of docs)

weixuanfu:
I just noticed that a new tool called [GAMA](https://github.com/PGijsbers/GAMA) recently released and it is developed by @PGijsbers. It is closely related to TPOT. The major differences are mentioned in [GAMA's docs](https://pgijsbers.github.io/gama/user_guide/index.html#tpot)


PGijsbers:
Thanks for the mention! :) Differences will probably become bigger over time, but the documentation should stay up to date.

Issue #877, opened 2019-06-03
sklearn/preprocessing/_function_transformer.py:98: FutureWarning: The default validate=True will be replaced by validate=False in 0.22


fcoppey:
Hello,

when I predict my test_set I get endless output of that :
/usr/local/lib/python3.7/site-packages/sklearn/preprocessing/_function_transformer.py:98: FutureWarning: The default validate=True will be replaced by validate=False in 0.22

any idea what to modify to make this go away?

thank you

fcoppey:
same problem with XGBoost :
WARNING: /workspace/src/objective/regression_obj.cu:152: reg:linear is now deprecated in favor of reg:squarederror.

something to change here too?

weixuanfu:
> Hello,
> 
> when I predict my test_set I get endless output of that :
> /usr/local/lib/python3.7/site-packages/sklearn/preprocessing/_function_transformer.py:98: FutureWarning: The default validate=True will be replaced by validate=False in 0.22
> 
> any idea what to modify to make this go away?
> 
> thank you


It seems FutureWarning message from the latest version of scikit-learn (0.21), we will update TPOT for fixing incompatibility issue in the next version of TPOT and should drop the supports of python 2.7 as scikit-learn 0.21 does. 


> same problem with XGBoost :
> WARNING: /workspace/src/objective/regression_obj.cu:152: reg:linear is now deprecated in favor of reg:squarederror.
> 
> something to change here too?

This seems from the API changes from [latest version of xgboost](https://github.com/dmlc/xgboost/releases). I am not sure how this warning message is triggered since TPOT does not set `reg:linear`. Maybe it is a bug in this version of xgboost. We will try to find a workaround for it.

fcoppey:
thank you! awesome work that you do guys! I wish I could help but I'm unfortunately not so useful with my knowledge yet hehe. maybe later...

shenglinqian:
got the same problem with XGBoost
WARNING: /workspace/src/objective/regression_obj.cu:152: reg:linear is now deprecated in favor of reg:squarederror
Anyone know how to figure it out?Thanks

thunfischtoast:
@shenglinqian @fcoppey you can prevent the deprecation warning by adding the following line to the config dictionary under the key ``xgboost.XGBRegressor``:

`` 'objective': ['reg:squarederror'] ``

If you don't have a dictionary yet just copy the default one from https://epistasislab.github.io/tpot/using/#built-in-tpot-configurations.

[...]

PR #887, opened 2019-07-11
Add explicit XGBoostRegressor objective to avoid deprecation error

thunfischtoast:
## What does this PR do?
Set the objective for XGBoostRegressors explicitly to reg:squarederror to prevent deprecation warnings.

tpot\tpot\config\regressor.py and tpot\tpot\config\regressor_sparse.py

From my point of view it it not needed.

Currently when including XGBoostRegressors in the pipeline the logs are spammed with deprecation warnings. 

Related partially to problems mentioned in #877 .

## Questions:


weixuanfu:
Good catch! Thank you!

PR #895, opened 2019-07-26
Adding support for Neural Network Embedding

### LINKED ISSUE #809: Add deep learning features to pipeline
Since the layer before the last layer in DNN have  important features extracted from origin input features. So, if you add the extracted features to auto_ml pipeline will enhance the whole performance of pipeline.

like as https://github.com/ClimbsRocks/auto_ml done.

NoRaincheck:

## What does this PR do?
Attempts to resolve #809 

Have a look at `tpot.builtins.embedding_estimator`

It has been tested on Python 3.7 on two environments; one with `tensorflow 2.0.0a` installed and one without `tensorflow` installed at all. It was tested via:

```
python3 -m nose tests/embedding_estimator_tests.py
```

See also #809 - as requested. I've made an attempt to ensure the API is compliant and sensible

#809 

## Questions:

- Do the docs need to be updated? Maybe - let me know what you need me to do.
- Does this PR add new (Python) dependencies? Adds new optional dependencies. You can use the MLPClassifier as part of existing dependencies (scikit-learn)

### DIFF
--- a/tests/embedding_estimator_tests.py
+++ b/tests/embedding_estimator_tests.py
@@ -1,0 +1,93 @@
+from sklearn.datasets import make_classification, make_regression
+from tpot.builtins import EmbeddingEstimator
+from sklearn.neural_network import MLPClassifier, MLPRegressor
+
+
+def test_EmbeddingClassifier_1():
+    """Assert that Embedding for classification works as expected."""
+    X, y = make_classification(random_state=1)
+    cs = EmbeddingEstimator(MLPClassifier(random_state=1, tol=0.9))
+    X_transformed = cs.fit_transform(X, y)
+
+    # 20 features + 100 embedding size
+    assert X_transformed.shape[1] == 120
+
+
+def test_EmbeddingClassifier_2():
+    """Assert that correct embedding layer is selected (classifier)."""
+    X, y = make_classification(random_state=1)
+    cs = EmbeddingEstimator(
+        MLPClassifier(hidden_layer_sizes=[20, 10], random_state=1, tol=0.9)
+    )
+    cs_2 = EmbeddingEstimator(
+        MLPClassifier(hidden_layer_sizes=[20, 10], random_state=1, tol=0.9),
+        embedding_layer=1,
+    )
+    X_transformed = cs.fit_transform(X, y)
+    X_transformed_2 = cs_2.fit_transform(X, y)
+
+    assert X_transformed.shape[1] == 30  # 20 features + 20 embedding size
+    assert X_transformed_2.shape[1] == 40  # 20 features + 20 embedding size
+
+
+def test_EmbeddingRegressor_1():
+    """Assert that Embedding for regressor works as expected."""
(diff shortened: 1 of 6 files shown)


tests/embedding_estimator_tests.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+def test_EmbeddingKeras():
+    """Check that this works also for keras models"""
+    try:
+        import tensorflow as tf
+    except ImportError:
+        tf = None
+    if tf is None:
+        return
```
NoRaincheck, line 67: Do you guys prefer raising a SkipTest here or leaving it like this?

tpot/builtins/embedding_estimator.py
```diff
@@ -0,0 +1,143 @@
...
+        X = check_array(X)
+        X_transformed = np.copy(X)
+        # add class probabilities as a synthetic feature
+        if issubclass(self.estimator.__class__, MLPClassifier) or issubclass(
+            self.estimator.__class__, MLPRegressor
+        ):
+            X_transformed = np.hstack(
+                (self._embedding_mlp(self.estimator, X), X_transformed)
+            )
+        else:
```
NoRaincheck, line 106: Chose not to do checking here otherwise you'll need to do an import at the top; which adds whatever flavour of the month neural network library to be a hard dependency to this. I don't have better ideas at this point but happy for any suggestions

tpot/builtins/embedding_estimator.py
```diff
@@ -0,0 +1,143 @@
...
+            The target values (integers that correspond to classes in classification, real numbers in regression).
+        fit_params:
+            Other estimator-specific parameters.
+
+        Returns
+        -------
+        self: object
+            Returns a copy of the estimator
+        """
+        if not issubclass(self.estimator.__class__, MLPClassifier) and not issubclass(
```
NoRaincheck, line 75: - not sure if this is the best way to check for Keras etc.

Q: how would we support pure tensorflow or pytorch? I've used Keras here to try to use the Keras scikit-learn wrapper

weixuanfu: Thank you for this PR. Keras scikit-learn wrapper is good so far.

tests/embedding_estimator_tests.py
weixuanfu: SkipTest should be OK since TF should be a optional dependency.

jhmenke:
is this still happening?

NoRaincheck:
Up to the maintainers

weixuanfu:
I am sorry for late response due to out of office for more than one month. I think removing  `MLPClassifier`/`MLPRegessor` (only using Keras wraper) in this transformer is more practical because  `MLPClassifier`/`MLPRegessor` are too slow on large dataset. As mentioned above, it will be an optional transformer. Also, it should be highly suggested on user guide that using Keras on GPU is highly recommended when using this transformer.

nickotto:
Please reimplement to the newest tpot implementation

Issue #836, opened 2019-02-19
XGBoost supports NaN but tpot enforces imputation


trhallam:
As a general rule tpot enforces imputation to match sklearn requirements for all real values in the input and output data. XGboost as a special case allows for the input of NaN values. 

## Context of the issue

I am trying to optimise XGboost specifically using a data set with quite a lot of holes in it. I do not want to perform imputation as it affects the results. I looked in base.py and quickly modified the `_check_data` function to ignore NaN values and to not perform imputation but was wondering if tpot can be modified to accommodate this scenario with XGboost?

A 'no_imputation' keyword might be added to `TPOTBase .__init__` for example to prevent imputation.

**Example Edits:**
```
else:
            if not self._imputed and np.any(np.isnan(features)):
                self._imputed = True
                features = self._impute_values(features)

        try:
            if target is not None:
                X, y = check_X_y(features, target, accept_sparse=True, dtype=np.float64, 

force_all_finite='allow-nan')

                return X, y
```

weixuanfu:
TPOT enforces imputation on dataset with NaN because most [operators in TPOT configuration](https://github.com/EpistasisLab/tpot/blob/master/tpot/config/classifier.py) do not support NaN. We may need another configuration if this `no_impuation` option is added.

trhallam:
I understand, it is a very specific case that I'm working on. Just currently there is no way to escape imputation with TPOT unless you modify the source. It is not a necessity perhaps more a nice to have.

Issue #771, opened 2018-09-25
How TPOT specify categorical feature by lgbm?


shuaianuoe:
Hi all,

As we know, lgbm is capable of specifying the categorical features through the **categorical_feature** parameter of the **fit** function. As follows：
`model = lgb.LGBMRegressor()
model .fit(X_train, y_train,feature_name=list(X_train.columns),
    categorical_feature=list(['col_1', 'col_2']))`

I know TPOT added **OneHotEncoder** to handle Categorical features. However,  the idea of ​​lgbm processing categories is not just based on one-hot. So, how can i use TPOT_CONFIG to  specify the categorical feature of lgbm?

Thank you very much and best regards!

weixuanfu:
The current version of TPOT cannot specify this parameters in `fit` function. But it is possible to expend [`set_sample_weight`](https://github.com/EpistasisLab/tpot/blob/v0.9.5/tpot/operator_utils.py#L81-L106) function to let TPOT `fit` function accept other additional keywords arguments as TPOT accepts `sample_weight`.

Issue #507, opened 2017-06-24
Add a built-in configuration dictionary for machine learning with text data


rhiever:
We can support machine learning with text data in TPOT by adding the [CountVectorizer](http://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.CountVectorizer.html) and [TfidfVecorizer](http://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.TfidfVectorizer.html#sklearn.feature_extraction.text.TfidfVectorizer) to a separate built-in configuration dictionary. I don't think we would need to change any of the other operators.

Unfortunately, without a pipeline grammar we can't force those vectorizers to always be at the beginning of every pipeline, but I suppose for text classification problems all of the pipelines that don't have one of the vectorizers will go "extinct."

Thoughts?

cc @weixuanfu2016 @teaearlgraycold

teaearlgraycold:
How would the text be passed to `fit`/`transform`? Right now we only handle float values for features, but both of these operators expect blocks of text.

weixuanfu2016:
Maybe we need a new text classification mode rather than a config file for transforming text input by the two operators.

ziarahman:
@weixuanfu2016, I like the idea of 'new text classification mode' .

@rhiever , thank you for opening this up for discussion. You mentioned CountVectorizer & TFIDFVectorizer. What about [HashingVectorizer](http://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.HashingVectorizer.html#sklearn.feature_extraction.text.HashingVectorizer)?

rhiever:
@teaearlgraycold, taking your example, it could work like this:

```python
my_text = "...".split("\n")
class_labels = [1, 0, 0, 1, 1, ..., 0, 0]
# Assuming len(my_text) == len(class_labels)

my_tpot = TPOTClassifier(..., config='TPOT text', ...)
my_tpot.fit(my_text, class_labels)
```

We would indeed need to change the dataset validation procedure when `config='TPOT text'`.

@ziarahman, I haven't used the HashingVectorizer. Does it take input data in the same format as the other two vectorizers?

weixuanfu2016:
how about `fit(features (non-text data), text, target)`? CountVectorizer & TFIDFVectorizer can be used on text data only and then stacking them in pipeline as below:

```
pipeline = make_pipeline(
    StackingText(CountVectorizer(input_text), input_matrix),
    TPOTClassifier()
)
```

[...]

Issue #889, opened 2019-07-11
sklearn's Imputer has been renamed & moved


rhiever:
As of sklearn v0.21.2 (possibly earlier), the `Imputer` has moved to the `impute` module and is now called the `SimpleImputer`: https://scikit-learn.org/stable/modules/generated/sklearn.impute.SimpleImputer.html

TPOT likely needs a quick update to update this reference when we perform imputation at the beginning of every TPOT run.

Also worth noting that the `impute` module offers [several different imputation methods](https://scikit-learn.org/stable/modules/classes.html#module-sklearn.impute) now, so perhaps it's worthwhile to move the imputation step into the configuration dictionary.

We can probably address #836 while making this change.

weixuanfu:
I agree with this update. Also, for addressing this and also supporting future version of sklearn (> 0.22), TPOT should not support Python 2 any more.

rhiever:
Agreed. It is time for future updates of TPOT to drop support for Python 2.

PR #897, opened 2019-07-27
Allow for Flexible Preprocessing

NoRaincheck:

## What does this PR do?

Too many things - so many things that I'm pretty sure there will be discussions on whether this is the way to proceed. Putting it up here before I invest more time on it. This PR addresses several items (possibly too many again) including

#507 
#836 
Handling of categorical data #771 - I know its not directly related; but it is the closest issue, and in my mind it would allow for extending it to using different encodings like using this: https://github.com/scikit-learn-contrib/categorical-encoding

Which are all related to how TPOT does preprocessing. This PR does a number of things including re-introducing "RandomTree" option in templates that allows specifying templates in the form `Transformer-RandomTree`; which will then allow things like `My_Preprocessing-RandomTree`.

The high level approach is to inject additional preprocessing steps when `_fit_init` is called which then alters the behaviour of TPOT. 

`tpot/base.py` - will add comments in files as part of this PR with my thoughts...

Also can't get relative imports working - so that would be appreciated for the `tpot.drivers.load_scoring_function` part

Happy to add new tests later; when the design is approved...

NIL see above

#507 
#836 
Handling of categorical data #771 

Example of API using modified Iris dataset

```py
import numpy as np
import pandas as pd

iris = load_iris()
    iris.target.astype(np.float64), train_size=0.75, test_size=0.25)

tpot = TPOTClassifier(generations=1, population_size=5, verbosity=2, template="PCA-RandomTree")

X_train_df = pd.DataFrame(X_train, columns=["num1", "num2", "num3", "num4"])
X_train_df['text'] = np.random.choice(["hello", "world", "foo", "bar world", "bar hello"], X_train.shape[0])

tpot2 = TPOTClassifier(generations=1, population_size=5, verbosity=2, template="PCA-LogisticRegression", 
                      preprocess_config_dict = {
                          'numeric_columns': ["num2"]
                      })
tpot2.fit(X_train_df, y_train)

tpot2 = TPOTClassifier(generations=1, population_size=5, verbosity=2, template="PCA-LogisticRegression", 
                      preprocess_config_dict = {
                          'numeric_columns': ["num2", "num3", "num4"]
                      })
tpot2.fit(X_train_df, y_train)

tpot2 = TPOTClassifier(generations=1, population_size=5, verbosity=2, template="PCA-LogisticRegression", 
                      preprocess_config_dict = {
                          'numeric_columns': ["num2"],
                          'text_columns': ['text']
                      })
tpot2.fit(X_train_df, y_train)
```

```
Generation 1 - Current best internal CV score: 0.5908385093167702

Best pipeline: LogisticRegression(PCA(PreprocessTransformer(input_matrix, numeric_columns=['num2']), iterated_power=2, svd_solver=randomized), C=20.0, dual=True, penalty=l2)

Generation 1 - Current best internal CV score: 0.9556935817805383

Best pipeline: LogisticRegression(PCA(PreprocessTransformer(input_matrix, numeric_columns=['num2', 'num3', 'num4']), iterated_power=9, svd_solver=randomized), C=20.0, dual=False, penalty=l1)

Generation 1 - Current best internal CV score: 0.5096273291925466
```

## Questions:

- Do the docs need to be updated? Yes - will do later

### DIFF
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -481,8 +481,11 @@ class TPOTBase(BaseEstimator):
 
 
     def _add_operators(self):
-        def add_randomtree():
-            step_in_type = np.ndarray
+        def add_randomtree(ret_types=[]):
+            if len(ret_types) > 0:
+                step_in_type = ret_types[-1]
+            else:
+                step_in_type = np.ndarray
             step_ret_type = Output_Array
             for operator in self.operators:
                 arg_types =  operator.parameter_types()[0][1:]
@@ -522,7 +525,7 @@ class TPOTBase(BaseEstimator):
                 if step == 'CombineDFs':
                     self._pset.addPrimitive(CombineDFs(), [step_in_type, step_in_type], step_in_type)
                 elif step == 'RandomTree':
-                    add_randomtree()
+                    add_randomtree(ret_types)
                 elif main_type.count(step): # if the step is a main type
                     for operator in self.operators:
                         arg_types =  operator.parameter_types()[0][1:]
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -65,7 +65,7 @@ from ._version import __version__
 from .operator_utils import TPOTOperatorClassFactory, Operator, ARGType
 from .export_utils import export_pipeline, expr_to_tree, generate_pipeline_code
 from .decorators import _pre_test
-from .builtins import CombineDFs, StackingEstimator
+from .builtins import CombineDFs, StackingEstimator, IdentityTransformer
 
 from .config.classifier_light import classifier_config_dict_light
 from .config.regressor_light import regressor_config_dict_light
@@ -110,8 +110,8 @@ class TPOTBase(BaseEstimator):
     def __init__(self, generations=100, population_size=100, offspring_size=None,
                  mutation_rate=0.9, crossover_rate=0.1,
                  scoring=None, cv=5, subsample=1.0, n_jobs=1,
-                 max_time_mins=None, max_eval_time_mins=5,
-                 random_state=None, config_dict=None, template=None,
+                 max_time_mins=None, max_eval_time_mins=5, random_state=None,
+                 config_dict=None, preprocess_config_dict=None, template=None,
                  warm_start=False, memory=None, use_dask=False,
                  periodic_checkpoint_folder=None, early_stop=None,
                  verbosity=0, disable_update_check=False):
@@ -210,6 +210,10 @@ class TPOTBase(BaseEstimator):
             String 'TPOT sparse':
                 TPOT uses a configuration dictionary with a one-hot-encoder and the
                 operators normally included in TPOT that also support sparse matrices.
+        preprocess_config_dict: a Python dictionary, optional (default: None)
+            Python dictionary:
+                A dictionary containing metadata on features for mandatory preprocessing.
+                Examples coming soon.
         template: string (default: None)
             Template of predefined pipeline structure. The option is for specifying a desired structure
             for the machine learning pipeline evaluated in TPOT. So far this option only supports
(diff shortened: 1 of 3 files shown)


tpot/builtins/preprocessing.py
```diff
@@ -0,0 +1,93 @@
...
+import_loader = {
+    'TfidfVectorizer': TfidfVectorizer,
+    'CountVectorizer': CountVectorizer,
+    'HashingVectorizer': HashingVectorizer,
+    'OneHotEncoder': OneHotEncoder
+}
+
+
+
+def load_scoring_function(scoring_func):
```
NoRaincheck, line 21: help for getting relative imports working would be appreciated here... `tpot.drivers.load_scoring_function`

tpot/base.py
```diff
@@ -391,6 +396,55 @@ def _setup_config(self, config_dict):
...
+                    if config_dict.get(k) is not None:
+                        column_transform_dict[k] = [config_dict[k]]
+                for k in ['text_transformer', 'categorical_transformer']:
+                    if k in config_dict:
+                        # force transformers to be a list
+                        if config_dict[k] is list:
+                            column_transform_dict[k] = config_dict[k]
+                        else:
+                            column_transform_dict[k] = [config_dict[k]]
+            self._config_dict['tpot.builtins.PreprocessTransformer'] = column_transform_dict
```
NoRaincheck, line 439: This injection could be dangerous - do we have opinions on how it is supposed to be handled?

tpot/base.py
```diff
@@ -391,6 +396,55 @@ def _setup_config(self, config_dict):
...
+
+            if len(check_keys_) > 0:
+                raise Exception(
+                    'It appears preprocessing configuration contains invalid keys: {}.'
+                    'Valid keys are text_columns, numeric_columns,'
+                    'categorical_columns, impute.'.format(', '.join(check_keys_))
+                )
+
+            # override some settings...
+            if config_dict.get('impute', None) is False:
```
NoRaincheck, line 424: This line deals with #836 - might be overloading this PR and might be an item for later?

tpot/base.py
weixuanfu: I think `preprocess_config_dict` should be a argument within PreprocessTransformer instead of TPOT. And users should be able to customize it via `config_dict`.

tpot/base.py
weixuanfu: I think it is more related to #889. I think we need add imputation into `config_dict` too. We may allow TPOT skip imputation if the pipeline only has XGBClassifier or XGBRegressor.

tpot/builtins/preprocessing.py
weixuanfu: Hmm, directly importing via `from ..driver import load_scoring_function` will cause some conflicts with 
```python
from .tpot import TPOTClassifier, TPOTRegressor
from ._version import __version__
```

A workaround is to move this function to `tpot/metrics.py` and then add `from ..metrics import load_scoring_function` to `tpot/builtins/preprocessing.py `

tpot/base.py
NoRaincheck: Yes this is certainly possible (and possible right now with no changes to TPOT master technically) via the use of templates - I think the question arises related to #507; if its possible to have a "built-in" configuration with text or not.

Maybe the answer is we can't

NoRaincheck:
Ok, 

Given the higher level comments @weixuanfu how should we proceed. In my opinion, I think this PR is just too big, what we probably want is:

*  a PR that just adds back "RandomTree" as an option so that we can do things like "PCA-RandomTree" or similar (have to ensure than RandomTree is the very last option"; in doing so this enables
*  Another PR that proposes how the preprocessing transform should work - there are downsides to this, as it means someone using the preprocessing has to directly alter config options for every model. e.g.

(as per comments above)

```
X_train, y_train = ...
my_meta_data_info = {<insert meta data information>}
config_dict = {**my_meta_data_info, **TPOT_DEFAULT_CONFIG}
tpot = TPOTClassifier(config_dict=config_dict, template='Preprocess-RandomTree')
tpot.fit(X_train, y_train)
```

versus

(as per approach taken in this PR)

```
X_train, y_train = ...
tpot = TPOTClassifier(config_dict=None, preprocess_config_dict ={<insert meta data>})
tpot.fit(X_train, y_train)
```

perib:
This is now possible with the new search space API in the latest version of TPOT

PR #903, opened 2019-08-14
Flag to allow multioutput.

mpdunne:

## What does this PR do?
Allows multi-output models to run with TPOT.

It's only a couple of small changes.

Try running TPOT with this flag. See if anything breaks.

Not all SK models support multiple regression. Therefore the user should also add MultiOutputRegressor in the config if they want to maximise the utility of the change.

https://github.com/EpistasisLab/tpot/issues/747
https://github.com/EpistasisLab/tpot/issues/810

## Questions:

- Do the docs need to be updated? - Yes.
- Does this PR add new (Python) dependencies? - No.

### DIFF
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -114,7 +114,7 @@ class TPOTBase(BaseEstimator):
                  random_state=None, config_dict=None, template=None,
                  warm_start=False, memory=None, use_dask=False,
                  periodic_checkpoint_folder=None, early_stop=None,
-                 verbosity=0, disable_update_check=False):
+                 verbosity=0, disable_update_check=False, multi_output=True):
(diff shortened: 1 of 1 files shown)


weixuanfu:
Thank you for this PR. The unit tests should pass but it failed due to installation of latest dask in CI environment. I will fix this issue in #904 and merge it to dev branch soon.

weixuanfu:
I am sorry for overlooking this issue for a while. I think a nicer solution is to let TPOT check y's shape to determine if TPOT is running on a multi-output problem and then switch configuration to Regressors which naively supports multi-output problems. Then the new parameter multi-output is not needed. Could you please make a update to this PR?

Issue #914, opened 2019-09-12
custom scorer does not work with n_jobs > 1 and use_dask=False


jhmenke:
A custom scorer works fine with n_jobs=1, but raises an Exeption with n_jobs > 1, because the scorer was not added to SCORERS

## Process to reproduce the issue

See above

## Expected result

The optimization continues as usual

## Current result

The new scorer is not added to SCORERS

## Possible fix

I would guess that due to the multiprocessing the SCORERS variable, where the new scorer is correctly added for the main process, the imports in the new processes refer to the original SCORERS in the metrics.py file

But i'm not sure how to elegantly solve this issue. A workaround is to statically add the custom scorer in metrics.py

weixuanfu:
Thank you for reporting this issue. I think maybe a better workaround is not to SCORERS for custom scorer ([related codes](https://github.com/EpistasisLab/tpot/blob/master/tpot/base.py#L320-L359)) but pass the scorer directly to [`_wrapped_cross_val_score`](https://github.com/EpistasisLab/tpot/blob/master/tpot/gp_deap.py#L382-L418). 

BTW, I am on vacation those days and will work on this issue later. But contribution is welcome.

jhmenke:
If it's okay to not user SCORERS in that case i might be able to fix it over the weekend.

PR #916, opened 2019-09-13
Enable custom scorer object for parallel computation #914

jhmenke:
Please see #914 

If a custom scorer is given as an object (not a str), it will be used directly instead of the SCORERS dict. Tests were adjusted accordingly. The behaviour for users does not change.


weixuanfu (approved):
Thank you for fixing this issue. I will merge this PR soon. I think we will refine scoring parameter (mentioned in that DeprecationWarning) a little bit more for next version of TPOT.

PR #920, opened 2019-09-27
Substitute cachedir with location

zarch:
## What does this PR do?

I saw in the Memory documentation that:

    'cachedir' has been deprecated in 0.12 and will be
    removed in 0.14. Use the 'location' parameter instead.
So I thought that it could be useful to make the change in `tpot`

## Questions:

No new dependencies but perhaps the `requirements.txt` file should change the line:
`joblib==0.10.3` into something like: `joblib>=0.12`.
I'v also noticed that the current line: `joblib=0.10.3` require an extra `=` to be valid.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,4 +7,4 @@ tqdm==4.26.0
 update-checker==0.16
 stopit==1.1.1
 pandas==0.20.2
-joblib=0.10.3
+joblib==0.10.3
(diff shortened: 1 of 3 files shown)


weixuanfu:
Thank you for catching this issue and this PR. I merged it to dev branch and will release the fix on next version of TPOT.

PR #939, opened 2019-10-26
added a boolean switch to export to screen instead of to file

### LINKED ISSUE #938: .export() to something else than a file
Sometimes running this library on hosted instances instead of local machines, the process of exporting to `.py` file is not always optimal. I might not have access to file output or they can get wiped after closing sessions.

An alternative way of reporting the best performing pipeline so far would be via stdout. For example, whenever a new best performing pipeline is found during the fitting process, a simple output of the `make_pipeline([...])` and its performance measure would suffice.

Is there something I am missing that already makes this possible? I could not find anything in the documentation in regards to the `.export()` function besides that [it takes a string as a parameter for the output file](https://epistasislab.github.io/tpot/api/#tpotclassifier-export).

gosuto-inzasheru:
## What does this PR do?
Fixes #938 by providing `.export()` with a new parameter; `to_screen`. If set to `True` export prints to screen instead of to file.

Try exporting to screen instead of to file!

## Discussion
I had to provide `output_file_name` with a default value. If not, a user would still have to provide a filename even when only exporting to screen.

## Documentation
Please see this as a suggestion. If agreed upon I can add some commits updating the docs before making this PR final (if needed).

### DIFF
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -1088,16 +1088,18 @@ class TPOTBase(BaseEstimator):
                 raise ValueError('Failed creating the periodic_checkpoint_folder:\n{}'.format(e))
 
 
-    def export(self, output_file_name, data_file_path=''):
+    def export(self, output_file_name='tpot_pipeline.py', data_file_path='', to_screen=False):
         """Export the optimized pipeline as Python code.
 
         Parameters
         ----------
-        output_file_name: string
+        output_file_name: string (default: 'tpot_pipeline.py')
             String containing the path and file name of the desired output file
         data_file_path: string (default: '')
             By default, the path of input dataset is 'PATH/TO/DATA/FILE' by default.
             If data_file_path is another string, the path will be replaced.
+        to_screen: boolean (default: False)
+            If set to True, the full text of the export is printed to screen instead of to file.
 
         Returns
         -------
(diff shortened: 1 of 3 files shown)


jhmenke:
Why is it exclusively to file OR to screen?

gosuto-inzasheru:
Fair point! I will adjust to AND/OR.

weixuanfu:
Please rebase your working branch to dev branch for avoiding conflicts.

tpot/base.py
```diff
@@ -1114,9 +1116,11 @@ def export(self, output_file_name, data_file_path=''):
                                     self.random_state,
                                     data_file_path=data_file_path)
 
-        with open(output_file_name, 'w') as output_file:
-            output_file.write(to_write)
-
+        if to_screen:
```
weixuanfu: Instead of printing the full text to stdout, maybe it is better to just return the string. User can easily print the returned string or grep information from it.

weixuanfu (changes requested):
Thank you for PR or issues. Please also add a unit tests into [`/tests/export_tests.py`] (https://github.com/EpistasisLab/tpot/blob/v0.10.2/tests/export_tests.py ) for this new parameters.

gosuto-inzasheru:
So I couldn't actually find any code that makes use of the returning of `True` or `False` based on whether the file was successfully written or not.

This PR therefore proposes to instead return the pipeline text as a string and only export to file when a file name is specified. In this way the function can be used to export to file AND/OR to string.

weixuanfu (approved):
Thanks again for refactoring this function. It looks good for merging now.

Issue #936, opened 2019-10-23
suggestion for more intuitive use of parameter max_time_mins


jhmenke:
Right now the logic is:

*generations* > 1 and *max_time_mins*=None --> tpot will run for *generations*
*generations* > 1 and *max_time_mins* > 0 --> tpot will run for ~*max_time_mins* minutes and ignore generations

However, if both *generations* and *max_time_mins* are defined, i would expect both to be a limiting factor. E.g, if generations is reached and the time is not up, stop regardless. The other way around, if generations is not reached but the time is up, tpot should be stopped.

What do you think about this proposal? I think it would be more intuitive if no parameter is just ignored.
That would also mean:

*generations*=None and *max_time_mins* > 0 --> tpot will run for *max_time_mins* minutes

weixuanfu:
I like this proposal and I think it's more logical.

weixuanfu:
The issue has been fixed in a new version of TPOT (v0.11.0). Please feel free to reopen it if you have any questions or suggestions.

PR #940, opened 2019-10-27
Max time mins fix

jhmenke:
Thix PR solves #936 by allowing to independently use generations and max_time_mins to limit the tpot search through using one of the parameters or both.


jhmenke:
Git messed up. Please don't merge yet, i will clean up my branch.

PR #941, opened 2019-10-27
allow both max_time_mins and generations as valid parameters concurre…

jhmenke:
Thix PR solves #936 by allowing to independently use generations and max_time_mins to limit the tpot search through using one of the parameters or both.

### DIFF
--- a/docs_sources/api.md
+++ b/docs_sources/api.md
@@ -34,7 +34,7 @@ Read more in the [User Guide](using/#tpot-with-code).
 <td width="80%" style="background:white;">
 <strong>generations</strong>: int, optional (default=100)
 <blockquote>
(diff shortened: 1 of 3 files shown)


jhmenke:
i updated the docstring as well

weixuanfu (changes requested):
Thank you for this PR.

Please also update the [docstring](https://github.com/EpistasisLab/tpot/blob/v0.10.2/tpot/base.py#L122-L126) and related documentation of generations parameter for the case of `generations=None`.

Also, for the case that `generations=None` AND `max_time_mins=None`, a `ValueError` should be raised for indicating invalid parameter combination. Then a unit test is needed for checking this `ValueError`.

jhmenke:
I made the appropriate changes.

weixuanfu (approved):

PR #972, opened 2019-12-03
There was a typo when linking to the portuguese bank example

nachopa:
There was a typo when linking to the portuguese bank example :-)

## What does this PR do?
It fixes the link to the portiguese bank example

It's a really small change, just the URL

Checking that the link works now

## Questions:


weixuanfu:
Thank you for the correction!

nachopa:
> Thank you for the correction!

Thank you for the project!  :-)

Issue #983, opened 2019-12-30
Scoring functions throw TPOTBase._update_top_pipeline() RuntimeError


jamesjjcondon:
## Context of the issue

'Accuracy' and 'balanced_accuracy' work fine, but others including f1, recall and precision
throw RuntimeError below

TPOT v11.0
sklearn 21.3

## Process to reproduce the issue

```
from tpot import TPOTClassifier
from sklearn.datasets import load_digits
from sklearn.model_selection import train_test_split
from sklearn.metrics.scorer import make_scorer

digits = load_digits()
X_train, X_test, y_train, y_test = train_test_split(digits.data, digits.target,
                                                    train_size=0.75, test_size=0.25, random_state=42)

tpot = TPOTClassifier(generations=5, population_size=20, verbosity=2,
                      scoring='f1', n_jobs=-1, random_state=42)

tpot.fit(X_train, y_train)
```

## Expected result

Score piped appropriately, 

## Current result

RuntimeError: There was an error in the TPOT optimization process. This could be because the data was not formatted properly, or because data for a regression problem was provided to the TPOTClassifier object. Please make sure you passed the data to TPOT correctly.

## Possible fix

Seems to be related to TPOTBase._update_top_pipeline(). When printing out pipeline and pipeline_scores from line 791 
https://github.com/EpistasisLab/tpot/blob/3d31727e6be5dc80b4dbf1e1835024e861b1eadf/tpot/base.py#L791 
score is `(1.0, -inf)`

Not sure about:
`self._optimized_pipeline_score = -float('inf')`

might be related to [868](https://github.com/EpistasisLab/tpot/issues/868)

Seems to happen regardless of n_jobs.

Any hints much appreciated. 

THANKYOU!!!!

weixuanfu:
The issue may be about using f1_score scorer in multi-class data, which is related to the issue #665. Please try “f1_macro” instead.

jamesjjcondon:
Of course! Thanks so much @weixuanfu . Yep, that was it, sorry. I thought my custom dataset was 2 class, but this made me realise there were nans that were forming a third class. I could do a PR to add to the error message?

jamesjjcondon:
Thanks so much @weixuanfu . Yep, that was it, sorry. I thought my custom dataset was 2 class, but this made me realise there were nans that were forming a third class. I could do a PR to add to the error message?

weixuanfu:
No problem! A PR for an user friendly error message is welcome.

[...]

PR #994, opened 2020-01-08
More user friendly error message

josiahjohnston:
Make an error message more user friendly for people who have errors in their custom scoring functions.

## What does this PR do?

Give the user another hint as to what may have caused an error.

Look at the diff

Standard tests

#983 

It would be nice to have an debugging option to make `tpot.gp_deap:_wrapped_cross_val_score()` print exceptions rather than hiding all exception. It would probably be best to control that via by the verbosity setting, but a separate hacky flag like `tpot.gp_deap.display_exceptions=True` would be adequate for local debugging purposes. 

#983 

## Questions:

No
No


weixuanfu:
Thank you for the PR and fix. I will merge it to dev branch.

Issue #971, opened 2019-12-03
multioutput problem


kradant:
I am working on a multioutput regression problem, that is the target values have more than 1-dim. 
A number of regressors from scikit-learn can only be used for multi output problems when used with the class MultiOutputRegressor (see especially https://scikit-learn.org/stable/modules/multiclass.html#multioutput-regression). 

MultiOutputRegressor takes a regressor as an argument and fits then one regressor per target. In this way most single-output regressors can deal multidimensional output. So I want to use it with TPOT. 

The issue  #903  deals with the changes that must be applied to base.py (which I did and this worked fine) in order to work for multiple output. But neither in #747, #810 nor in #903 it was clarified how to actually use MultiOutputRegressor with **several** regressors. 

This is my config dictionary:
`custom_regressor_config_dict = {
    'sklearn.multioutput.MultiOutputRegressor': {
        'estimator': 
            {'sklearn.linear_model.ElasticNetCV': {
                'l1_ratio': np.arange(0.0, 1.01, 0.05),
                'tol': [1e-5, 1e-4, 1e-3, 1e-2, 1e-1]}}}`

Now I want to insert **more** regressors. But I tried to put them in a list or in dictionary and whatever, for example I tried

`'sklearn.multioutput.MultiOutputRegressor': {
        'estimator': [
            {'sklearn.linear_model.ElasticNetCV': {
                'l1_ratio': np.arange(0.0, 1.01, 0.05),
                'tol': [1e-5, 1e-4, 1e-3, 1e-2, 1e-1]}},
            {'sklearn.ensemble.AdaBoostRegressor': {
                'n_estimators': [100],
                'learning_rate': [1e-3, 1e-2, 1e-1, 0.5, 1.],
                'loss': ["linear", "square", "exponential"]}}
        ]            
    }
`

 and it always throws an error like: 

> RuntimeError: There was an error in the TPOT optimization process. This could be because the data was not formatted properly[...]

How to format it properly so that I can use more models?  Is there a workaround? Thank you!

weixuanfu:
TPOT currently do not support multi-output regression and current configuration did not support more than one `estimator` option (similar to #956) within this kind of meta estimators, but we are working on adding supports for > 1 estimators and I think that is the first step for supporting multi-output regression.

kradant:
Do you know an alternative tool like TPOT that can do multi-output regression? Or a workaround? Like starting a loop with TPOT and changing each time the estimator?

[...]

PR #1001, opened 2020-01-14
automatic detection of multioutput datasets

jhmenke:
## What does this PR do?

Slight variation of #903 depending on the shape of the target array.

#971


weixuanfu:
Neat! Thank you for this PR. I think it should work for checking multi output dataset. But I checked this [link](https://stats.stackexchange.com/questions/153853/regression-with-scikit-learn-with-multiple-outputs-svr-or-gbm-possible), I think that not all regressors in our default config are supporting multi-output regression. I think we need a workaround for this issue.

jhmenke:
There are several ways to go about this:
- Give a warning if the standard regression / classification config is chosen for multi outputs, which would require the user to create their own config for multi output datasets.
- Make a list with regressors and classifiers that don't support multi output and remove them automatically if such a dataset is detected
- Embed these regressors and classifiers in MultiOutputRegressor/Classifier in the default config. Not sure about the performance losses there

weixuanfu:
Thank you for your ideas here. I prefer the 3rd one but I think a better solution based on it is to automatically use `MultiOutputRegressor/Classifier` over pipelines that generated from current default config if TPOT detected the y is a multi-output target. Please let me know if you have any idea.

jhmenke:
While working on it, i think i spotted a bug:

https://github.com/EpistasisLab/tpot/blob/aea42a59b52ff9704f9f677a01ebb71fda4520cc/tpot/operator_utils.py#L215


This should be *dep_import_str* as the key and not *import_str*, right? It seems this also changes some tests.

jhmenke:
If i run the failing tests manually, they work. It has to be some unsuitable combination due to the random seed.. i'm not sure how to debug this.

weixuanfu:
> 
> 
> While working on it, i think i spotted a bug:
> 
> https://github.com/EpistasisLab/tpot/blob/aea42a59b52ff9704f9f677a01ebb71fda4520cc/tpot/operator_utils.py#L215
> 
> This should be _dep_import_str_ as the key and not _import_str_, right? It seems this also changes some tests.

Yes, that is a bug. Hmm, I thought I fixed it a while ago but might not merge to master/dev branch.

tpot/base.py
```diff
@@ -501,6 +500,35 @@ def _fit_init(self):
             self._last_optimized_pareto_front_n_gens = 0
             self._setup_config(self.config_dict)
 
+            if multi_output_target:
```
weixuanfu, line 509: Modifying _config_dict may not work in the situation that use use a customized configurations  instead of default one. So, I think a practical way is to modify the `_compile_to_sklearn` function ([here](https://github.com/EpistasisLab/tpot/blob/v0.11.1/tpot/base.py#L1193-L1194)). If `multi_output_target` is `True`, then
`sklearn_pipeline=MultiOutputClassifier(estimator=sklearn_pipeline)` or  `sklearn_pipeline=MultiOutputRegessor(estimator=sklearn_pipeline)` . I think it maybe a more general solution for multioutput dataset.

jhmenke: This seems better, to be honest i didn't find a good place where to put my code and only settled on the _fit_init function therefore.

jhmenke: Okay i looked into it, but the code would be a mess. Several functions would have to take multi_output_target as a new argument (most of them in export_utils.py), since they don't have access to the data or the TPOT Object. 

Imo _fit_init seems to be the least intrusive point to include the checks

weixuanfu: Thank you for looking into. You are right. I think TPOT exported codes should also include MultiOutputRegessor/MultiOutputClassifier, which should change a lot of codes in TPOT. I will look into it when I get some time next week.

jhmenke: any updates?

weixuanfu: @jhmenke I am sorry for overlooking this. I did not get a chance to look into this issue those days due to my busy schedule. I agree that TPOT need some major changes for including `MultiOutputRegessor/MultiOutputClassifier`. I may get some time in March to add those changes. You are welcome to push any changes meanwhile.

jhmenke: Could we use my PR as a temporary fix until you have time to thoroughly refactor the code? I can prepare an update with the current development branch.

weixuanfu: Sorry for the delay. I think we can use it for a temporary solution with a minor release.

jhmenke: Okay i merged the current master/development into this. Should be good to go as an interim solution then.

tpot/base.py
```diff
@@ -507,6 +506,35 @@ def _fit_init(self):
...
+                    'sklearn.linear_model.ElasticNetCV',
+                    'sklearn.svm.LinearSVR',
+                    'xgboost.XGBRegressor',
+                    'sklearn.linear_model.SGDRegressor'
+                ]
+                for model in list(self._config_dict.keys()):
+                    if model in single_output_classifiers:
+                        if 'sklearn.multioutput.MultiOutputClassifier' not in self._config_dict.keys():
+                            self._config_dict['sklearn.multioutput.MultiOutputClassifier'] = {"estimator": {}}
+                        self._config_dict['sklearn.multioutput.MultiOutputClassifier']['estimator'][model] = self._config_dict[model]
```
weixuanfu, line 527: There is only one sklearn.multioutput.MultiOutputClassifier in self._config_dict and `estimator` is keeping updated until the last model in  `single_output_classifiers` so that the rest models should be removed.

weixuanfu (changes requested):
I think we cannot add this support for now since it need more changes.

jessegmeyerlab:
Is this the closest pull to multi-output? any plans to finish it?

jhmenke:
> 
> 
> Is this the closest pull to multi-output? any plans to finish it?

sorry, no time for this right now..  feel free to work with my branch. i think the last todo is cloning the multioutputclassifier for every estimator that only supports single outputs as per weixuanfu's review.

windowshopr:
Would love to see multioutput regression/classification working with tpot! Keep going! Haha

PR #1004, opened 2020-01-16
Bugfix stratify pretest sample

kristiankaufmann:
## What does this PR do?

Fixes an edge case that arises when using tpot on dataset with extreme class imbalance where the pretest sample would cause an error to be thrown since the sample resulted in only one class being represented in the pretest_sample

See the changes in the tpot base with the introduction of a init_pretest function and the corresponding implementation in the TPotClassifier

a unit test was added to verify that the pretest function works as intended

see PR description
## Questions:

- Do the docs need to be updated? no
- Does this PR add new (Python) dependencies? no

### DIFF
--- a/tpot/tpot.py
+++ b/tpot/tpot.py
@@ -23,6 +23,11 @@ License along with TPOT. If not, see <http://www.gnu.org/licenses/>.
 
 """
 
+from sklearn.model_selection import train_test_split
+from sklearn.utils import _safe_indexing
+
+import numpy as np
+
 from .base import TPOTBase
 from .config.classifier import classifier_config_dict
 from .config.regressor import regressor_config_dict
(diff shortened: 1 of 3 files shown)


tpot/tpot.py
```diff
@@ -35,7 +40,17 @@ class TPOTClassifier(TPOTBase):
     default_config_dict = classifier_config_dict  # Classification dictionary
     classification = True
     regression = False
+    def _init_pretest(self, features,target):
+        """Set the sample of data used to verify pipelines work with the passed data set.
+
+        This is not intend for anything other than perfunctory dataset pipeline compatibility testing
+        """
 
+        self.pretest_X, _, self.pretest_y, _ = train_test_split(features, target, random_state=self.random_state,
```
weixuanfu, line 49: Could you please use `stratify` in `train_test_split` function for this classification? Then maybe the checking below is not needed.

weixuanfu (changes requested):
Thank you for finding this issue and submitting this PR. Just one suggestion below for train/test splits in classification. Maybe we could also use that `stratify` param in `train_test_split` for regression. If so, I think the fix will be a little simpler.

tpot/tpot.py
kristiankaufmann: I attempted to use the stratify, but as I discovered it preserves the proportions of the classes. Thus if the sample size is limited to 50 samples and the class proportions more imbalanced than 50:1 then passing the stratify parameter to test train split fails the solve the issue.

weixuanfu: Hmm, thank you for the test. I think we will use the solution in this PR after some test.

weixuanfu:
Thank you for the PR. I merge this patch for a temp fix.

PR #1018, opened 2020-02-18
Closes #1017. Fix pickling of CV generators.

### LINKED ISSUE #1017: Pickling failure for custom CV splitter
Working:
```python3
from sklearn.model_selection import LeaveOneGroupOut
import numpy as np


X = np.array([[1, 2], [3, 4], [5, 6], [7, 8]])
y = np.array([1, 2, 1, 2])
groups = np.array([1, 1, 2, 2])

logo = list(LeaveOneGroupOut().split(X=X, y=y, groups=groups))

t = TPOTClassifier(verbosity=2, max_time_mins=1, n_jobs=-1, cv=logo)
t.fit(features=X, target=y)
```

Not working:
```python3
logo = LeaveOneGroupOut().split(X=X, y=y, groups=groups)

t = TPOTClassifier(verbosity=2, max_time_mins=1, n_jobs=-1, cv=logo)
t.fit(features=X, target=y)
```
```
TypeError: can't pickle generator objects
_pickle.PicklingError: Could not pickle the task to send it to the workers.
```

It seems like the `logo` generator is getting sent to `joblib`

Looking at the `sklearn` source, it seems like the generator gets converted into an array before parallelizing (see [here](https://github.com/scikit-learn/scikit-learn/blob/21686b7147ea9f24a8e415d05a5046477e976864/sklearn/model_selection/_validation.py#L735)). I tried stealing `check_cv` from `sklearn` and it fixed the issue, I'm not sure about importing a function from a `_` prefixed module in `sklearn`, otherwise I think this would be the ideal solution.

adriangb:
Fixes pickling of custom cv split generators to fix bug described in #1017 

I ended up having to make changes to a couple of fo the existing tests because they interacted with `_wrapped_cross_val_score` directly and that API has now changed. The general API for `fit` and `TPOT` instantiation is unchanged. I also added a test for this bug.

I do not think any of the docs need to be updated, please correct me if I am wrong.

### DIFF
--- a/tests/tpot_tests.py
+++ b/tests/tpot_tests.py
@@ -54,7 +54,7 @@ from tempfile import mkdtemp
 from shutil import rmtree
 
 from sklearn.datasets import load_digits, load_boston, make_classification
-from sklearn.model_selection import train_test_split, cross_val_score, GroupKFold
+from sklearn import model_selection
 from joblib import Memory
 from sklearn.metrics import make_scorer, roc_auc_score
 from sklearn.base import BaseEstimator, ClassifierMixin, RegressorMixin, TransformerMixin
@@ -67,7 +67,7 @@ from deap.tools import ParetoFront
 from nose.tools import nottest, assert_raises, assert_not_equal, assert_greater_equal, assert_equal, assert_in
 from driver_tests import captured_output
 
-train_test_split = nottest(train_test_split)
+train_test_split = nottest(model_selection.train_test_split)
 
 with warnings.catch_warnings():
     warnings.simplefilter('ignore')
@@ -103,8 +103,21 @@ training_features_r, testing_features_r, training_target_r, testing_target_r = \
 # Set up a small test dataset
 
 pretest_X, pretest_y = make_classification(n_samples=100, n_features=10, random_state=42)
-# Set up pandas DataFrame for testing
 
+# Set up artificial groups
+n_datapoints = pretest_y.shape[0]
+groups = np.random.randint(low=0, high=n_datapoints//2, size=n_datapoints)
+
+# Set up custom cv split generators
+custom_cvs = [
+    model_selection.LeaveOneGroupOut().split(X=pretest_X, y=pretest_y, groups=groups),
+    model_selection.LeavePGroupsOut(2).split(X=pretest_X, y=pretest_y, groups=groups),
+    model_selection.RepeatedStratifiedKFold(),
+    5,
+    10,
+    ]
+
+# Set up pandas DataFrame for testing
 input_data = pd.read_csv(
     'tests/tests.csv',
     sep=',',
@@ -335,16 +348,38 @@ def test_timeout():
     tpot_obj._optimized_pipeline = creator.Individual.from_string(pipeline_string, tpot_obj._pset)
     tpot_obj.fitted_pipeline_ = tpot_obj._toolbox.compile(expr=tpot_obj._optimized_pipeline)
     # test _wrapped_cross_val_score with cv=20 so that it is impossible to finish in 1 second
+    cv = model_selection.KFold(n_splits=20).split(
+        X=training_features_r,
+        y=training_target_r,
+    )
+    cv = model_selection._split.check_cv(cv, training_target_r, classifier=False)
     return_value = _wrapped_cross_val_score(tpot_obj.fitted_pipeline_,
                                             training_features_r,
                                             training_target_r,
-                                            cv=20,
+                                            cv=cv,
                                             scoring_function='neg_mean_squared_error',
                                             sample_weight=None,
                                             groups=None,
                                             timeout=1)
     assert return_value == "Timeout"
 
+def test_custom_cv_test_generator():
+    """Check that custom cv generators processed correctly.
+    """
+    def check_custom_cv(_cv):
+        tpot_obj = TPOTClassifier(
+            random_state=42,
+            population_size=1,
+            offspring_size=2,
+            generations=1,
+            verbosity=0,
+            cv=_cv,
+            n_jobs=-1  # ensure pickling / parallelization
+        )
+        tpot_obj.fit(pretest_X, pretest_y)
+
+    for cv in custom_cvs:
+        yield check_custom_cv, cv
 
 def test_invalid_pipeline():
     """Assert that _wrapped_cross_val_score return -float(\'inf\') with a invalid_pipeline"""
@@ -357,11 +392,20 @@ def test_invalid_pipeline():
     )
     tpot_obj._optimized_pipeline = creator.Individual.from_string(pipeline_string, tpot_obj._pset)
     tpot_obj.fitted_pipeline_ = tpot_obj._toolbox.compile(expr=tpot_obj._optimized_pipeline)
-    # test _wrapped_cross_val_score with cv=20 so that it is impossible to finish in 1 second
+
+    cv = model_selection.KFold().split(
+        X=training_features,
+        y=training_target,
+    )
+    cv = model_selection._split.check_cv(
+        cv,
+        training_target,
+        classifier=False  # choice of classifier vs. regressor here is arbitrary
+    )
     return_value = _wrapped_cross_val_score(tpot_obj.fitted_pipeline_,
                                             training_features,
                                             training_target,
-                                            cv=5,
+                                            cv=cv,
                                             scoring_function='accuracy',
                                             sample_weight=None,
                                             groups=None,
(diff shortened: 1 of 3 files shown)


tests/tpot_tests.py
```diff
@@ -54,7 +54,7 @@
 from shutil import rmtree
 
 from sklearn.datasets import load_digits, load_boston, make_classification
-from sklearn.model_selection import train_test_split, cross_val_score, GroupKFold
+from sklearn import model_selection
```
adriangb, line 57: Since I was calling a couple more classes in `model_selection`, I just changed the import to import the entire module.

tests/tpot_tests.py
```diff
@@ -103,8 +103,21 @@ def closing(arg):
...
-# Set up pandas DataFrame for testing
 
+# Set up artificial groups
+n_datapoints = pretest_y.shape[0]
+groups = np.random.randint(low=0, high=n_datapoints//2, size=n_datapoints)
+
+# Set up custom cv split generators
+custom_cvs = [
+    model_selection.LeaveOneGroupOut().split(X=pretest_X, y=pretest_y, groups=groups),
+    # model_selection.LeavePGroupsOut(2).split(X=pretest_X, y=pretest_y, groups=groups),
```
adriangb: Ups realized I forgot to uncomment these. I will uncomment and update after others review.

tests/tpot_tests.py
```diff
@@ -335,16 +348,38 @@ def test_timeout():
...
+    cv = model_selection.KFold(n_splits=20).split(
+        X=training_features_r,
+        y=training_target_r,
+    )
+    cv = model_selection._split.check_cv(cv, training_target_r, classifier=False)
     return_value = _wrapped_cross_val_score(tpot_obj.fitted_pipeline_,
                                             training_features_r,
                                             training_target_r,
-                                            cv=20,
+                                            cv=cv,
```
adriangb, line 359: This is necessary because `_wrapped_cross_val_score` will no longer accept an `int` as it's cv parameter. We should consider renaming the `cv` parameter to `cv_splitter` or something to make this distinction more verbose, but that might require a lot of small changes everywhere, I'm not sure if it's worth it.

tests/tpot_tests.py
```diff
@@ -335,16 +348,38 @@ def test_timeout():
...
                                             scoring_function='neg_mean_squared_error',
                                             sample_weight=None,
                                             groups=None,
                                             timeout=1)
     assert return_value == "Timeout"
 
+def test_custom_cv_test_generator():
```
adriangb, line 366: New test for this specific bug. Not very fancy, just pass a couple of custom cv generators and make sure there are no errors.

tests/tpot_tests.py
```diff
@@ -357,11 +392,20 @@ def test_invalid_pipeline():
     )
     tpot_obj._optimized_pipeline = creator.Individual.from_string(pipeline_string, tpot_obj._pset)
     tpot_obj.fitted_pipeline_ = tpot_obj._toolbox.compile(expr=tpot_obj._optimized_pipeline)
-    # test _wrapped_cross_val_score with cv=20 so that it is impossible to finish in 1 second
```
adriangb, line 360: This comment was inaccurate so I removed it.

tests/tpot_tests.py
weixuanfu: OK, looks good.

tests/tpot_tests.py
weixuanfu: Please uncomment it.

tests/tpot_tests.py
weixuanfu: I think keeping `cv` is ok. Please also update docstring around [this line](https://github.com/EpistasisLab/tpot/blob/v0.11.1/tpot/gp_deap.py#L397)

tests/tpot_tests.py
weixuanfu: OK, it looks good.

tests/tpot_tests.py
weixuanfu: Thank you for cleaning it.

weixuanfu (approved):
Thank you for making those changes. Please update a little based on my comments then I will merge it to dev branch once I get a chance.

tests/tpot_tests.py
adriangb: Doesn't this change make the docstring _valid_? Before this change, using a cross-validation generator did not work. Now it does. I'm sorry if I am missing something.

weixuanfu: I think int won't be accepted after this changes. So the docstring need remove `Int` and `If CV is a number, then it is the number of folds to evaluate each` from the `_wrapped_cross_val_score`

adriangb: I'm very sorry. I followed the link and thought I was looking at the documentation for the `TPOT` class instantiation. You are right, the behavior for `_wrapped_cross_val_score` changes. I will update the docstring.

adriangb: I'm thinking of changing the documentation to reflect that `cv` is expected to be an instance of a generator whos items support the `split` method with the signature `split(X, y, groups)`. Something like this:

```
cv: cross-validation split generator instance
        Instance of a generator that that gives the train/test of each fold
        via the ``split`` method with signature ``split(X, y, groups)``.
```

Does this make sense?

weixuanfu: Hmm, we want cv param in TPOT to be consistent with cv param in cross_eval_score from scikit-learn to avoid misunderstanding. So I think maybe this new way is not needed for now although I think it is a good idea.

adriangb: Okay, I will make the docstring simply:
```
cv: cross-validation generator
        Object to be used as a cross-validation generator.
```

adriangb:
@weixuanfu should be ready to merge 👍

weixuanfu:
Thank you a lot! Merged!

adriangb:
Awesome!





Glad I could help!

I didn't want to include it in this PR, but I did want to bring this up: is there any reason for the `cv` parameter to be in the `TPOT` constructor? It seems like it's not really used until `fit` is called.

weixuanfu:
The reason is that we don’t want fit() to take params which don’t follow scikit-learn API.

adriangb:
That makes sense. What about making a `TPOT.cross_validate` method that essentially achieves the same thing as the current implementation, but abstracts it out to a seperate function to keep the API closer to sklearn?

weixuanfu:
Hmm, I think that do not follow scikit-learn API for Estimator neither. Also, those cv splits is for evaluating those pipelines in TPOT with average cv scores (as one of fitness scores in GP) which has different purpose with `cross_validate` from scikit-learn that can compute/store all cv scores and even fitted models on all splits

adriangb:
You're right, thank you for entertaining the thought.

Maybe a quick summary of my use case will be better at explaining why I was asking. My workflow is as follows (pseudocode but you get the idea):
```python3
# setup
raw_data = pd.DataFrame(...)
feature_extractors = [ext_ft_1, ext_ft_2, ...]   # callable methods ext_ft(raw_data) -> X, y, groups
models = [
     TPOTClassifier(max_time_mins=10),
     KNeighborsClassifier(),
]

# extract featues -> fit -> score
for ft_ext in feature_extractors:
    X, y, groups = ft_ext(raw_data)
    cv = list(LeaveOneGroupOut().split(X, y, groups))
    for model in models:
        model.fit(X, y)
        if isinstance(model, TPOTClassifier):
             model =model.fitted_pipeline_
       score = np.mean(cross_val_score(estimator=model, X=X, y=y, cv=cv)
    # save model, score
```

This works with the `scikit-learn` models, but not with `TPOT`. So I modified, kind of as follows:
```python3
# setup
raw_data = pd.DataFrame(...)
feature_extractors = [ext_ft_1, ext_ft_2, ...]
models = [
    (TPOTClassifier, {"max_time_mins": 10,}),
    (KNeighborsClassifier, dict()),
]

# extract featues -> fit -> score
for ft_ext in feature_extractors:
    X, y, groups = ft_ext(raw_data)
    cv = list(LeaveOneGroupOut().split(X, y, groups))
    for model_class, opts in models:
        if isinstance(model, TPOTClassifier):
             model = model_class(**opts, cv=cv)
             model.fit(X, y)
             model =model.fitted_pipeline_
       else:
            model = model_class(**opts)
            model.fit(X, y)
       score = np.mean(cross_val_score(estimator=model, X=X, y=y, cv=cv)
      # save model, score
```

I know this is a contrived use case, but I was just wondering if there is a better way.

weixuanfu:
But for the original workflow, how about adding the cv into the TPOTClassifier’s param setting? Or for modified workflow how about using [model.set_params](https://scikit-learn.org/stable/modules/generated/sklearn.base.BaseEstimator.html#sklearn.base.BaseEstimator.set_params). I think both follow scikit-learn API.

adriangb:
Using `model.set_params` worked beautifully! I wasn't aware of that API. Thank you.

PR #1020, opened 2020-02-19
Fix string literal comparison (wrong operator)

ckastner:
Two strings are being compared with 'is not', when they should be compared with the != operator.


weixuanfu:
Thank you for the PR. I will merge it to dev branch.

Issue #1023, opened 2020-02-21
Long evaluation in _pre_test hangs search


josepablocam:
Thanks for the great tool. When using it, I ran into a bit of
an edge case. For some operators, the
reasonable `_pre_test` decorator (`tpot/decorators.py`)
can actually take an arbitrarily long amount of time, effectively
hanging the search (since it never moves past generating candidates).
The small sampled down `pretest_X` might be reasonable, but
for some operators complexity scales with columns (and it may be
hard to determine the correct number to sample, if any).

You can reproduce the issue by using the following example:
```
> cat example.py

import tpot
import numpy as np
import pmlb

X, y = pmlb.fetch_data("Hill_Valley_without_noise", return_X_y=True)

clf = tpot.TPOTClassifier(
  max_time_mins=1,
  max_eval_time_mins=1,
  cv=2,
  config_dict={
      'sklearn.decomposition.SparsePCA':{},
      'sklearn.linear_model.LogisticRegression':{},
  },
  verbosity=3,
)

clf.fit(X, y)
```

This will timeout, rather than finish appropriately

```
timeout 5m python example.py # will hang (and get killed 5 minutes after)
echo $? # 124 -> timeout
```

## Possible fix

I have a simple patch that adds a timeout to the
`.fit` call taking place at line 88 in `tpot/decorators.py`. I can PR it if this seems useful.

weixuanfu:
Hmm, nice suggestion! Please submit a PR for this issue. Thank you!

PR #1024, opened 2020-02-22
Fix `_pre_test` hanging when generating candidates

josepablocam:
This PR addresses the issue #1023

We added a time out to the pipeline `.fit` call.

Running the below on the original codebase should hang, while the patched version should complete as expected.

```
> cat example.py

import tpot
import numpy as np
import pmlb

X, y = pmlb.fetch_data("Hill_Valley_without_noise", return_X_y=True)

clf = tpot.TPOTClassifier(
  max_time_mins=1,
  max_eval_time_mins=1,
  cv=2,
  config_dict={
      'sklearn.decomposition.SparsePCA':{},
      'sklearn.linear_model.LogisticRegression':{},
  },
  verbosity=3,
)

clf.fit(X, y)
```

```
timeout 5m python example.py
echo $?
```

Note that I chose a timeout of 2 seconds in the patch (an arbitrary choice). I leave the choice of this to the maintainers, who are likely to have a better sense of a reasonable value.

## Questions:

No

No

### DIFF
--- a/tpot/decorators.py
+++ b/tpot/decorators.py
@@ -29,7 +29,12 @@ import warnings
 from .export_utils import expr_to_tree, generate_pipeline_code
 from deap import creator
 
+from stopit import threading_timeoutable, TimeoutException
+
+
 NUM_TESTS = 10
+MAX_EVAL_SECS = 2
+
 
 def _pre_test(func):
     """Check if the wrapped function works with a pretest data set.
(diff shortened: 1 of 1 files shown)


weixuanfu:
Thank you for the PR and it  looks good for fixing the issue #1023. I will merge it to dev branch for now and add this patch to next version of TPOT.

PR #1025, opened 2020-02-24
Deep learning

adriangb:
## What does this PR do?
Attempting to revive #895 while expanding upon it a little to allow for more general integration of `Keras` models into `TPOT`.

The main focus is basic sequential models (`tensorflow.keras.models.Sequential`).

This PR proposes a method of wrapping them to allow parametric generation of deep learning models in `TPOT`. Example classes are created for a classifier, a regressor and a transformer (the latter being the original intent of #895).

This is a _very_ rough proof of concept, but I believe that the versatile functionality of `tensorflow.keras` will allow for fancy stuff like callbacks to be integrated into TPOT's logging, etc.

Please start by looking over the general implementation and API in `deep_learning.py`.
I realize that extensive documentation is needed, I will do that at some point, I'm hoping to get _some_ feedback before I spend a bunch of time on that.

I wrote some basic tests. More is needed.

I also _think_ I found a bug in `tensorflow.keras.utils.generic_utils.has_arg` which would be a blocker for this. I submitted a PR (tensorflow/tensorflow#37004) on the `tensorflow` repo, we'll see where that goes. If tests on this PR fail, that's probably why.

I also ran some basic tests comparing the performance of these parametrically generated estimators with some basic Keras tutorials I found, it's about the same if not better.

#895 , #809

PS: @chappers, I made a new PR because it's a new branch on my fork. I squashed your commits into a single commit for now, I will make sure you are a co-author if this gets merged.


tests/test_deep_learning.py
```diff
@@ -0,0 +1,38 @@
+import nose
+from sklearn.datasets import make_classification, make_regression
+from sklearn.neural_network import MLPClassifier, MLPRegressor
```
weixuanfu, line 3: Please remove this line since this two seems not to be used in the scripts.

tpot/builtins/deep_learning.py
```diff
@@ -0,0 +1,223 @@
...
+        activation=output_layer_activation,
+        kernel_regularizer=kernel_regularizer,
+        kernel_initializer=kernel_initializer,
+    ))
+    model.compile(
+        optimizer=optimizer, loss=loss, metrics=metrics
+    )
+    return model
+
+class DeepLearningClassifier(KerasClassifier):
```
weixuanfu, line 98: Based on the error [in AppVayor](https://ci.appveyor.com/project/weixuanfu/tpot/builds/31010281), it failed to import tpot due to no tensorflow in the environment. Maybe using `HAS_TENSORFLOW` to determine if those new classes should be created.

tests/test_deep_learning.py
```diff
@@ -0,0 +1,38 @@
...
+if not HAS_TENSORFLOW:
+    raise nose.SkipTest()
+
+
+def test_Embedding_Keras():
+    """Assert that Embedding for classification works as expecated."""
+    layer_sizes = [20, 100, 50, 20, 60, 100]
+    X, y = make_classification(random_state=1)
+
+    def check(X, X_transformed, embedding_layer_size):
```
weixuanfu, line 19: This function is duplicated and please only keep one.

weixuanfu (changes requested):
Thank you for submitting this PR. I like this feature and I think it should be merged into TPOT if the performance in computational time and accuracy/score is competitive with other scikit-learn estimators in default TPOT configuration (e.g vs. RandomForestEstimator). Please provide more details about the performance test (with/without GPU) and also the search space of hyperparameters. Regarding the callbacks function from `tensorflow.keras`, it can be integrated later.

tests/test_deep_learning.py
weixuanfu: Pleas also fix those failed unit tests in this test scripts.

adriangb:
Great, thank you for taking a look and pointing some things out. I'll do the following three things as next steps:

1. Fix CI issues
2. Fix comments made above
3. Compare performance and accuracy. The primary comparison will be to `sklearn.neaural_net.MLPClassifier`. I will compare accuracy to other models, but I fully expect this to be a lot slower than classic classifiers. I will probably use [this](https://scikit-learn.org/stable/auto_examples/classification/plot_classifier_comparison.html) scikit-learn doc as a basepoint for comparison.


Quick update: I'm realizing that in order to do a 'fair' comparison, I'm going to need to tune all of the default parameters for these networks to as similar as possible to `MPLClassifier`. From initial testing with this dataset, these networks can meet or exceed the accuracy of `MPLClassifier`, but the runtime is considerably worse. This is all using `tensorflow` in CPU only on WSL. I'm sure using the GPU version on bare metal would already make this at least as fast as `MPLClassifier`, but I think tuning those parameters as I mentioned should allow the CPU version to be about as fast.

adriangb:
Another update. Although I _was_ able to hack this together to make a `DeepLearningRegressor` class with the exact same API as `MPLClassifier`, there are some bugs and it uses somewhat ugly mixin inheritance patterns. Looking around the `tf.keras` source, I found this PR, which I think would make things a lot cleaner and easier on the `TPOT` side: tensorflow/tensorflow#32533

Hence, I think I am going to focus efforts on reviving that PR and will pause work on this one until then.

adriangb:
Another comment. Testing performance, I found that for small datasets on my laptops CPU, tensorflow/keras was _considerably_ slower than `sklearn.neural_net`. I think running on any decent GPU, distributed or maybe even a workstation CPU, tensorflow would be much faster, but I really don't have the ability to test that.

adriangb:
If anyone is interested, this could easily be revived using https://github.com/adriangb/scikeras. MLP example: https://github.com/adriangb/scikeras/blob/master/tests/mlp_models.py

PR #1031, opened 2020-03-15
adjusted error message

jamesjjcondon:
## What does this PR do?
Includes example of common error source

n/a

n/a

I came across this when first using TPOT with binary classes, binary-only evaluation metrics and nans that were treated as a third class. This is likely a common use-case and error for beginners.
(my first contribution : )

## Questions:


weixuanfu:
Thank you for rewording!

PR #1038, opened 2020-04-03
Add parameter "progress file"

Jaimecclin:

## What does this PR do?

Add a new parameter to output tqdm progress to a file.

tpot/base.py

1. Test on progress_file as None
2. Test on progress_file as file_handler

## Questions:

api.md is already updated.
Nope.


tpot/base.py
```diff
@@ -232,7 +233,8 @@ def __init__(self, generations=100, population_size=100, offspring_size=None,
             A setting of 2 or higher will add a progress bar during the optimization procedure.
         disable_update_check: bool, optional (default: False)
             Flag indicating whether the TPOT version checker should be disabled.
-
+        progress_file: io.TextIOWrapper or io.StringIO, optional (defaul: sys.stdout)
+            Save progress content to a file.
```
weixuanfu, line 240: Could you please rename the `progress_file` to `log_file`? I think there are some log infos when `verbosity=3`.

tpot/base.py
```diff
@@ -263,6 +265,7 @@ def __init__(self, generations=100, population_size=100, offspring_size=None,
         self.verbosity = verbosity
         self.disable_update_check = disable_update_check
         self.random_state = random_state
+        self._file = progress_file
```
weixuanfu: Based on scikit-learn API, please also use `self.log_file = log_file` in __init__.

weixuanfu (changes requested):
Thank you for the PR. Could you please provide small examples (with population=20, generations=10) of this log file when verbosity=1, 2, 3?

Jaimecclin:
> Thank you for the PR. Could you please provide small examples (with population=20, generations=10) of this log file when verbosity=1, 2, 3?

Of course I can. I am wondering that I should write an example and take screenshots or write an unit test. How do you think?

weixuanfu:
unit tests are preferred.

Jaimecclin:
Hi,
I added a file to test verbosity from 1-3. In case of verbosity as 3, an error was raised. It seems it uses xgboost when verbosity is 3. 

`
  File "/app/tpot/operator_utils.py", line 80, in source_decode
    exec('from {} import {}'.format(import_str, op_str))
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'xgboost'
`

I am not sure the reason so I just commented out the last unit test. And it also happens in master branch. @weixuanfu Could you please help to answer this? If it's an issue, I can try to figure out it.

Jaimecclin:
> Hi,
> I added a file to test verbosity from 1-3. In case of verbosity as 3, an error was raised. It seems it uses xgboost when verbosity is 3.
> 
> `Traceback (most recent call last): File "/app/tpot/operator_utils.py", line 80, in source_decode exec('from {} import {}'.format(import_str, op_str)) File "<string>", line 1, in <module> ModuleNotFoundError: No module named 'xgboost'`
> 
> I am not sure the reason so I just commented out the last unit test. And it also happens in master branch. @weixuanfu Could you please help to answer this? If it's an issue, I can try to figure out it.

Hi,

I found the reason why it raise exception only in case of verbosity as 3. 

```
    tmp_path = sourcecode.split('.')
    op_str = tmp_path.pop()
    import_str = '.'.join(tmp_path)
    try:
        if sourcecode.startswith('tpot.'):
            exec('from {} import {}'.format(import_str[4:], op_str))
        else:
            exec('from {} import {}'.format(import_str, op_str))
        op_obj = eval(op_str)
    except Exception as e:
        if verbose > 2:
            raise ImportError('Error: could not import {}.\n{}'.format(sourcecode, e))
        else:
            print('Warning: {} is not available and will not be used by TPOT.'.format(sourcecode))
        op_obj = None

    return import_str, op_str, op_obj
```
I think it can be modified as printing warning simply instead of raising error. I can put my unit test 3 back as well. Any suggestion is welcome. :)

weixuanfu:
Thank you for looking into it.

The issue is that there is no xgboost in the test environment. We intentionally let tpot raise that import error when verbosity=3. 

I think in the CI envs of TPOT unit tests ([Linux](https://github.com/EpistasisLab/tpot/blob/master/ci/.travis_install.sh#L41) and [Windows](https://github.com/EpistasisLab/tpot/blob/master/.appveyor.yml#L23)), xgboost should be installed so that this error should not be raised. 

Could you please build a similar conda env with xgboost for testing unit test with verbosity=3?

Jaimecclin:
Ok. I tried it and passed all tests. Please take a look. Thanks.

tests/test_progress_parameter.py
```diff
@@ -0,0 +1,72 @@
+# -*- coding: utf-8 -*-
```
weixuanfu: Please change the unit test scripts name to `test_log_file.py`

tpot/base.py
```diff
@@ -263,6 +265,7 @@ def __init__(self, generations=100, population_size=100, offspring_size=None,
         self.verbosity = verbosity
         self.disable_update_check = disable_update_check
         self.random_state = random_state
+        self._file = log_file
```
weixuanfu: Based on [scikit-learn API for BaseEstimator](https://scikit-learn.org/stable/modules/generated/sklearn.base.BaseEstimator.html#sklearn.base.BaseEstimator), the line should be self.log_file=log_file. Then TPOT object can both use `get_params` and `set_params`.

weixuanfu (changes requested):
Thank you for updating. I just found a few minor farther changes. I will merge it to dev branch first for next version of TPOT once those changes are made.

tests/test_progress_parameter.py
Jaimecclin: Done. It's a better naming. Thanks.

tpot/base.py
Jaimecclin: That makes a lot sense. It was finished. Thanks.

Jaimecclin:
Hi,
Should I resolve that conflict in base.py?

weixuanfu:
> Hi,
> Should I resolve that conflict in base.py?

 I have resolved it. Thank you!

weixuanfu:
Thank you for the PR. I will merge it to dev branch.

Jaimecclin:
Thanks for your help, @weixuanfu. :)

PR #1073, opened 2020-05-29
Write generation info to log_file in high verbosity mode

adrienhenry:
## What does this PR do?
Unsure that the log system also logs the generation informations to the log file when the later is provided.
Currently, the generation information are printed to the stdout even with a log_file 

- updates in gp_deap.py
- updates in test_log_file.py

It uses the existing suite

I also reduced the number of generations and pop_size  in test_log_file.
The quality of the prediction is not important for the tests. Reducing the these parameters saves up some time on the tests.

https://github.com/EpistasisLab/tpot/issues/1072

## Questions:


weixuanfu:
Thank you for the PR!

PR #1076, opened 2020-06-01
Make `torch` an optional dependency

### LINKED ISSUE #1075: New undeclared dependency on pytorch?
Congratulations on the release!It looks like as of `0.11.4`, importing `tpot` is trying to bring in `torch`.

While I see it was added to `requirements.txt`, it would need adding to `install_requires` in `setup.py`, as `pip` doesn't really look at anything else but that file (well, and `setup.cfg` and `pyproject.toml`... but i digress...)

**But**, to be frank, there's a whole raft of complexity that comes along with a hard dependency on torch, and as a downstream maintainer, and occasional instructor of using `tpot`, I'd selfishly **strongly** prefer it be hidden behind an `extra` like some of the other heavyweight dependencies (e.g. `xgboost`, `dask`) unless it's _really_ part of the core `tpot` functionality now. 

Indeed, to that end, as conda-forge only ships `pytorch-cpu`, and I generally can't recommend using `pytorch-cpu` (from conda-forge, [pytorch](https://anaconda.org/pytorch) or otherwise) for getting any real work done (other than testing), the documentation would likely need to be updated to suggest using `-c pytorch -c defaults -c conda-forge tpot pytorch` to get a useful setup... but then, I haven't been able to test that hypothesis yet! At any rate, we _would_ be able to support a hard pin, with [`run_constrained`](https://docs.conda.io/projects/conda-build/en/latest/resources/define-metadata.html#run-constrained)... but that can be an unsatisfying and bewildering user experience if they didn't even _know_ `tpot` used `torch` until this patch release!

## Context of the issue

Downstream pull request:
- https://github.com/conda-forge/tpot-feedstock/pull/25#issuecomment-636078946

CI screesnshot (with link to log):
[Screenshot from 2020-05-29 20-18-06](https://dev.azure.com/conda-forge/feedstock-builds/_build/results?buildId=168375&view=logs&j=1bf226d3-0e2f-52d8-fa93-7d9e633347b3&t=ec8d466d-e3b4-5115-4f06-222398f91e6c&l=719
)

JDRomano2:
## What does this PR do?
Fixed #1075 by making PyTorch an optional dependency. TPOT will only load NN modules if `torch` is available to import. If the user tries to directly import TPOT-NN estimators (e.g., `from tpot.builtins import nn`) without PyTorch installed, an appropriate error message will be raised.

`nosetests -s -v --with-coverage`

n/a

#1075 

n/a

## Questions:

Yes, new information added to installation instruction Markdown source.
No, a dependency was removed.

### DIFF
--- a/docs_sources/installing.md
+++ b/docs_sources/installing.md
@@ -58,6 +58,10 @@ If you plan to use the [TPOT-MDR configuration](https://arxiv.org/abs/1702.01780
 pip install scikit-mdr skrebate
 ```
 
+To enable support for [PyTorch](https://pytorch.org/)-based neural networks (TPOT-NN), you will need to install PyTorch. TPOT-NN will work with either CPU or GPU PyTorch, but we strongly recommend using a GPU version, if possible, as CPU PyTorch models tend to train very slowly.
+
+We recommend following [PyTorch's installation instructions](https://pytorch.org/get-started/locally/) customized for your operating system and Python distribution.
+
 Finally to install TPOT itself, run the following command:
 
 ```Shell
--- a/optional-requirements.txt
+++ b/optional-requirements.txt
@@ -1,3 +1,4 @@
 xgboost==0.90
 scikit-mdr==0.4.4
 skrebate==0.3.4
+torch==1.3.1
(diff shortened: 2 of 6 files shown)


weixuanfu (commented):
Thank you for the quick fix. The updated information added to installation instruction Markdown source seems not committed yet. Please add updated info about installation of torch as optional package.

PR #1084, opened 2020-06-15
fix export line in NN example

cjber:
Changed tpot.export to clf.export as the code uses clf rather than tpot naming in above examples.


weixuanfu:
Thank you! Nice catch!

PR #1109, opened 2020-08-29
Ignore npnanmean runtimewarning

gosuto-inzasheru:
The `np.nanmean` used for meaning the CV scores throws a `RuntimeWarning: Mean of empty slice` when the `CV_score` array consists of nothing else than `np.nan`. 

The error message gets in the way of the general progress output and it could be argued that [the error message is non-sensical to start with](https://stackoverflow.com/questions/29688168/mean-nanmean-and-warning-mean-of-empty-slice).

I believe the `warnings.catch_warnings()` is implemented here for exactly that reason: to suppress these warnings. The last `return` line however also can throw this error and should therefore be indented as well.

### DIFF
--- a/tpot/gp_deap.py
+++ b/tpot/gp_deap.py
@@ -447,7 +447,7 @@ def _wrapped_cross_val_score(sklearn_pipeline, features, target,
     else:
         try:
             with warnings.catch_warnings():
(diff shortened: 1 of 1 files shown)


weixuanfu:
Thank you for the fix!

PR #1136, opened 2020-11-09
Added Descriptive Error on Bad `optype`

AyrtonB:


AyrtonB:
I had two errors with CI but neither are related to the changes in this PR

```
======================================================================
FAIL: Assert that TPOT template option generates pipeline when one of steps is a specific operator.
----------------------------------------------------------------------
  File "C:\Miniconda36-x64\envs\test-environment\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\projects\tpot\tests\tpot_tests.py", line 779, in test_template_3
    assert sklearn_pipeline.steps[0][0] == 'SelectPercentile'.lower()
AssertionError
======================================================================
FAIL: Assert that TPOT operators return their type, e.g. 'Classifier', 'Preprocessor'.
----------------------------------------------------------------------
  File "C:\Miniconda36-x64\envs\test-environment\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\projects\tpot\tests\tpot_tests.py", line 2445, in test_operator_type
    assert TPOTSelectPercentile.type() == "Selector"
AssertionError
----------------------------------------------------------------------

```

tpot/operator_utils.py
```diff
@@ -194,11 +194,12 @@ def TPOTOperatorClassFactory(opsourse, opdict, BaseClass=Operator, ArgBaseClass=
         elif is_regressor(op_obj):
             class_profile['root'] = True
             optype = "Regressor"
-        elif _is_selector(op_obj):
-            optype = "Selector"
         elif _is_transformer(op_obj):
             optype = "Transformer"
-
+        elif _is_selector(op_obj):
```
weixuanfu, line 207: `_is_selector` need be moved above `_is_transformer` because "Selector" actually is a special kind of "Transformer". If the logic checks "Transformer" first, then there is no Selector in default configuration. It will fix that 2 failed unit tests.

weixuanfu (changes requested):

AyrtonB:
now included in [PR-1137](https://github.com/EpistasisLab/tpot/pull/1137)

PR #1137, opened 2020-11-10
Added Resampler Components

AyrtonB:
## What does this PR do?

Enables the inclusion of imblearn resampler components to the TPOT pipeline

This builds on [issue 457](https://github.com/EpistasisLab/tpot/issues/547), the detailed changes are discussed in [this comment](https://github.com/EpistasisLab/tpot/issues/547#issuecomment-723973738)

Through inclusion of imblearn components

```python
from sklearn.datasets import make_classification

classifier_config_dict = {
    # Classifiers
    'sklearn.ensemble.ExtraTreesClassifier': {
        'n_estimators': [100],
        'criterion': ["gini", "entropy"],
        'max_features': np.arange(0.05, 1.01, 0.05),
        'min_samples_split': range(2, 21),
        'min_samples_leaf': range(1, 21),
        'bootstrap': [True, False]
    },

    # Preprocessors
    'imblearn.over_sampling.RandomOverSampler': {
    },
    
}

X, y = make_classification(n_samples=5000, n_features=2, n_informative=2,
                           n_redundant=0, n_repeated=0, n_classes=3,
                           n_clusters_per_class=1,
                           weights=[0.01, 0.05, 0.94],
                           class_sep=0.8, random_state=0)

pipeline_optimizer = TPOTClassifier(generations=5, population_size=10, cv=3,
                                   random_state=42, verbosity=2, n_jobs=-1, 
                                   config_dict=classifier_config_dict, 
                                   template='RandomOverSampler-Classifier')

pipeline_optimizer.fit(X, y)

pipeline = pipeline_optimizer.fitted_pipeline_
```

Output
```
Generation 1 - Current best internal CV score: 0.98940007916784

Generation 2 - Current best internal CV score: 0.98940007916784

Generation 3 - Current best internal CV score: 0.9896000391758383

Generation 4 - Current best internal CV score: 0.9896000391758383

Generation 5 - Current best internal CV score: 0.9897999991838367

Best pipeline: ExtraTreesClassifier(RandomOverSampler(input_matrix), bootstrap=False, criterion=gini, max_features=0.3, min_samples_leaf=1, min_samples_split=10, n_estimators=100)
Wall time: 28.4 s
```

### DIFF
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -569,8 +569,7 @@ class TPOTBase(BaseEstimator):
             )
         self._toolbox.register("mutate", self._random_mutation_operator)
 
-    def _fit_init(self):
-        # initialization for fit function
+    def _get_make_pipeline_func(self):
         imblearn_used = (
             Series(self.config_dict.keys())
             .str.split(".")
@@ -586,6 +585,11 @@ class TPOTBase(BaseEstimator):
         else:
             make_pipeline_func = make_pipeline
 
+        return make_pipeline_func
+
+    def _fit_init(self):
+        # initialization for fit function
+
         if not self.warm_start or not hasattr(self, "_pareto_front"):
             self._pop = []
             self._pareto_front = None
@@ -597,6 +601,9 @@ class TPOTBase(BaseEstimator):
 
             self.operators = []
             self.arguments = []
+
+            make_pipeline_func = self._get_make_pipeline_func()
+
             for key in sorted(self._config_dict.keys()):
                 op_class, arg_types = TPOTOperatorClassFactory(
                     key,
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -571,7 +571,7 @@ class TPOTBase(BaseEstimator):
 
     def _get_make_pipeline_func(self):
         imblearn_used = (
-            Series(self.config_dict.keys())
+            Series(self._config_dict.keys())
             .str.split(".")
             .str[0]
             .isin(["imblearn"])
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -48,7 +48,8 @@ from copy import copy, deepcopy
 
 from sklearn.base import BaseEstimator
 from sklearn.utils import check_X_y, check_consistent_length, check_array
-from sklearn.pipeline import make_pipeline, make_union
+from sklearn.pipeline import make_union
+from imblearn.pipeline import make_pipeline
 from sklearn.preprocessing import FunctionTransformer
 from sklearn.impute import SimpleImputer
 from sklearn.model_selection import train_test_split
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -49,7 +49,6 @@ from copy import copy, deepcopy
 from sklearn.base import BaseEstimator
 from sklearn.utils import check_X_y, check_consistent_length, check_array
 from sklearn.pipeline import make_union, make_pipeline
-from imblearn.pipeline import make_pipeline as make_imblearn_pipeline
 from sklearn.preprocessing import FunctionTransformer
 from sklearn.impute import SimpleImputer
 from sklearn.model_selection import train_test_split
@@ -88,6 +87,11 @@ from .gp_deap import (
     _wrapped_cross_val_score,
     cxOnePoint,
 )
+
+try:
+    from imblearn.pipeline import make_pipeline as make_imblearn_pipeline
+except:
+    make_imblearn_pipeline = None
 
 with warnings.catch_warnings():
     warnings.simplefilter("ignore")
@@ -577,6 +581,7 @@ class TPOTBase(BaseEstimator):
         )
 
         if imblearn_used == True:
+            assert make_imblearn_pipeline is not None, "You must install `imblearn`"
             make_pipeline_func = make_imblearn_pipeline
         else:
             make_pipeline_func = make_pipeline
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -40,7 +40,7 @@ from tempfile import mkdtemp
 from shutil import rmtree
 
 import numpy as np
-from pandas import DataFrame
+from pandas import DataFrame, Series
 from scipy import sparse
 import deap
 from deap import base, creator, tools, gp
@@ -48,8 +48,8 @@ from copy import copy, deepcopy
 
 from sklearn.base import BaseEstimator
 from sklearn.utils import check_X_y, check_consistent_length, check_array
-from sklearn.pipeline import make_union
-from imblearn.pipeline import make_pipeline
+from sklearn.pipeline import make_union, make_pipeline
+from imblearn.pipeline import make_pipeline as make_imblearn_pipeline
 from sklearn.preprocessing import FunctionTransformer
 from sklearn.impute import SimpleImputer
 from sklearn.model_selection import train_test_split
@@ -61,7 +61,12 @@ from update_checker import update_check
 
 from ._version import __version__
 from .operator_utils import TPOTOperatorClassFactory, Operator, ARGType
-from .export_utils import export_pipeline, expr_to_tree, generate_pipeline_code, set_param_recursive
+from .export_utils import (
+    export_pipeline,
+    expr_to_tree,
+    generate_pipeline_code,
+    set_param_recursive,
+)
 from .decorators import _pre_test
 from .builtins import CombineDFs, StackingEstimator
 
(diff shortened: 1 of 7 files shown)


AyrtonB:
It's failing on `ModuleNotFoundError: No module named 'imblearn'` but I added imblearn to the requirements.txt

AyrtonB:
also added `imblearn` to `setup.py` but I'm still getting `ModuleNotFoundError: No module named 'imblearn'`

weixuanfu:
Thank you for the PR. Please add installation of `imblearn` into [ci environment in travis](https://github.com/EpistasisLab/tpot/blob/master/ci/.travis_install.sh) and [appveyor](https://github.com/EpistasisLab/tpot/blob/master/.appveyor.yml) to fix the `ModuleNotFoundError` . This PR seems very similar with #1136. Could you please merge them into one PR?

AyrtonB:
Have now added them to both, for some reason the AppVeyor build was cancelled. Thank you for helping with this.

weixuanfu:
The PR did not referred a correction Appveyor build, the correct one is [here][https://ci.appveyor.com/project/weixuanfu/tpot/builds/36233033]. `ModuleNotFoundError` is gone, but we got other errors. I think the error is from changing import source of `make_pipeline` from `sklearn` to `imblearn`.  Can those samplers in imblearn work with sklearn's `make_pipeline`?

I think we need make imblearn as optional dependency of TPOT.

AyrtonB:
Unfortunately the samplers require the `imblearn` `Pipeline`

> [The difficulty here is that ImbLearn applies `fit` and `sample`, notice the latter is not `transform` as it does not change features (transformations), only the re-samples (hence sampling). For this reason, ImbLearn provides its own `Pipeline` module, as it needs to wrap the `sample` functionality in a way that makes sense (it only samples on training and not on testing, etc) and is compatible with SciKit-Learn API flow.](https://github.com/EpistasisLab/tpot/issues/547#issuecomment-451115193) - [@romanovzky](https://github.com/romanovzky)  

How would you recommend best integrating it as an optional dependency of TPOT?

weixuanfu:
How about adding checks whether a configuration `config_dict` includes imblearn or not. If so, TPOT and exported pipeline should use make_pipeline from imblearn.

AyrtonB:
I've made it toggleable on whether an `imblearn` component is specified in the `config_dict` for `base.py`, however I also need to change it in `export_utils.py` under `_starting_imports` which `config_dict` isn't being passed to.

tpot/base.py
```diff
@@ -40,16 +40,16 @@
...
 import deap
 from deap import base, creator, tools, gp
 from copy import copy, deepcopy
 
 from sklearn.base import BaseEstimator
 from sklearn.utils import check_X_y, check_consistent_length, check_array
-from sklearn.pipeline import make_union
-from imblearn.pipeline import make_pipeline
+from sklearn.pipeline import make_union, make_pipeline
+from imblearn.pipeline import make_pipeline as make_imblearn_pipeline
```
weixuanfu: Please use `try...except...` for environment without `imblearn`

weixuanfu:
> I've made it toggleable on whether an `imblearn` component is specified in the `config_dict` for `base.py`, however I also need to change it in `export_utils.py` under `_starting_imports` which `config_dict` isn't being passed to.

Yep, for the pipeline did not involve `imblearn`, then `make_pipeline` from imblearn is not needed.

tpot/base.py
AyrtonB: now added

AyrtonB:
Thank you. Would I need to change anything here then or could I leave it as sklearn?

```
    if num_op_root > 1:
        return {
            "sklearn.model_selection": ["train_test_split"],
            "sklearn.pipeline": ["make_union", "make_pipeline"],
            "tpot.builtins": ["StackingEstimator"],
        }
    elif num_op > 1:
        return {
            "sklearn.model_selection": ["train_test_split"],
            "sklearn.pipeline": ["make_pipeline"],
        }
```

weixuanfu:
> Thank you. Would I need to change anything here then or could I leave it as sklearn?
> 
> ```
>     if num_op_root > 1:
>         return {
>             "sklearn.model_selection": ["train_test_split"],
>             "sklearn.pipeline": ["make_union", "make_pipeline"],
>             "tpot.builtins": ["StackingEstimator"],
>         }
>     elif num_op > 1:
>         return {
>             "sklearn.model_selection": ["train_test_split"],
>             "sklearn.pipeline": ["make_pipeline"],
>         }
> ```

Yes, if a pipeline includes imblearn, then it should use make_pipeline from imblearn. If not, then use the one from sklearn. I hope sklearn can support/include imblearn in the future.

tpot/base.py
```diff
@@ -498,27 +537,60 @@ def _add_terminals(self, arg_types):
...
-            self._toolbox.register('expr_mut', self._gen_grow_safe, min_=self._min, max_=self._max)
-        self._toolbox.register('mutate', self._random_mutation_operator)
+            self._toolbox.register(
+                "expr_mut", self._gen_grow_safe, min_=self._min, max_=self._max
+            )
+        self._toolbox.register("mutate", self._random_mutation_operator)
+
+    def _get_make_pipeline_func(self):
+        imblearn_used = (
+            Series(self.config_dict.keys())
```
weixuanfu: self.config_dict is None by default, please use `_config_dict` instead after calling [`self._setup_config(self.config_dict)`](https://github.com/EpistasisLab/tpot/blob/v0.11.6.post1/tpot/base.py#L526)

AyrtonB (commented):
`_get_make_pipeline_func` is called after `self._setup_config(self.config_dict)` is run so I'm confused as to why it's still None


Edit:
Saw review on refresh, now changing to `self._config_dict`

weixuanfu:
> `_get_make_pipeline_func` is called after `self._setup_config(self.config_dict)` is run so I'm confused as to why it's still None

Please check my comment above.

Just using `self._config_dict` instead

tpot/export_utils.py
```diff
@@ -21,6 +21,13 @@
 
 import deap
 
+try:
+    import imblearn
+
+    pipeline_module = "imblearn"
```
weixuanfu: This is not the proper way to determine if a pipeline includes imblearn's resampler. I think you can check `op.import_hash` after this [line](https://github.com/EpistasisLab/tpot/blob/v0.11.6.post1/tpot/export_utils.py#L203) to determine if the pipeline includes a resampler from 'imblearn'. If so, put this [line](https://github.com/EpistasisLab/tpot/blob/v0.11.6.post1/tpot/export_utils.py#L200) after the check and then use imblearn's make_pipeline instead sklearn's one. It may require one new argument about the check for the `_starting_imports ` function.

AyrtonB: It should work as a generalised way of checking to see if they have imblearn and if so using that pipeline, I will change to only use imblearn on requriement in pipeline though. 

In `import_relations` are the values just strings of the library names?

AyrtonB:
There are now just 3 tests failing 🎉 but I can't work out how to get past them. 

They all seem to be related to the exporting of the pipeline

```
======================================================================
FAIL: Assert that generate_import_code() returns the correct set of dependancies and dependancies are importable.
----------------------------------------------------------------------
  File "C:\Miniconda36-x64\envs\test-environment\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\projects\tpot\tests\export_tests.py", line 277, in test_generate_import_code_2
    assert expected_code == import_code
AssertionError
======================================================================
FAIL: Assert that exported_pipeline() generated a compile source file as expected given a fixed pipeline.
----------------------------------------------------------------------
  File "C:\Miniconda36-x64\envs\test-environment\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\projects\tpot\tests\export_tests.py", line 340, in test_export_pipeline
    assert expected_code == export_pipeline(pipeline, tpot_obj.operators, tpot_obj._pset)
AssertionError
======================================================================
FAIL: Assert that exported_pipeline() generated a compile source file as expected given a fixed simple pipeline with input_matrix in CombineDFs.
----------------------------------------------------------------------
  File "C:\Miniconda36-x64\envs\test-environment\lib\site-packages\nose\case.py", line 197, in runTest
    self.test(*self.arg)
  File "C:\projects\tpot\tests\export_tests.py", line 448, in test_export_pipeline_4
    assert expected_code == export_pipeline(pipeline, tpot_obj.operators, tpot_obj._pset)
AssertionError
```

weixuanfu:
I merged this PR to development branch with a commit https://github.com/EpistasisLab/tpot/commit/0716a34719d2942682ce3294c3fe1ca0a3576178 that should fixed the 3 tests above.

weixuanfu:
I think I need add more unit tests with config_dict that includes imblearn resampler before merging the changes to master branch. Please feel free to submit another PR if you want to add unit tests.

AyrtonB:
Thank you very much. I'll have to wait a little bit but will have a shot at them

Issue #1110, opened 2020-09-05
Preprocessing on specific features


hanshupe:
I see that after the TPOT optimization a preprocessor like robustScaler was selected. I wonder if it's possible that robustScaler is applied not on the entire set of features but only on a few were it makes sense. One feature may have outliers and require robustScaler(0.1, 0.9), but the other features not. Can this be considered with TPOT?

weixuanfu:
I think current TPOT may not support this kind of application. But I think it need adding the [ColumnTransformer](https://scikit-learn.org/stable/modules/generated/sklearn.compose.ColumnTransformer.html) from scikit-learn into TPOT.

edwardyu:
Interested in working on this. `ColumnTransformer` takes a list of columns. How to make this list available to the genetic algorithm?

edwardyu:
OK I have an idea. In `TPOTBase.fit()`, add a hook called `_dynamically_modify_config_dict()`, which will modify `config_dict` with a dict like this:

```
    'tpot.builtins.ColumnTransformer': {
        'transformers': [''sklearn.preprocessing.StandardScaler', 'sklearn.preprocessing.RobustScaler', ...],
        'include_col_1': [True, False],
        'include_col_2': [True, False],
        ...
        'include_col_n': [True, False],
    },
```

What do you think @weixuanfu ?

Issue #1078, opened 2020-06-09
Using TPOT's GP framework to select features


sokol11:
Hi. I wonder if it's possible to use TPOT's genetic programming framework to select the best subset of features. I already have an idea of what classification algorithm and parameters work well and I do not want to optimize the entire ML pipeline. I just would like to create a population of smaller feature sets out of ~1000 features I have and use cross-over/mutation to find the best set while using my static estimator for model evaluation. Can I do that with TPOT? If so, how might I go about doing it? Thanks!

weixuanfu:
TPOT has a built-in FeatureSetSelector (see [this link](https://epistasislab.github.io/tpot/using/#featuresetselector-in-tpot)) for helping user to select best feature set based on _priori_ expert knowledge. Also you may limit ML hyperparameter space in `config_dict` (like using a static estimator in the dictionary) and fix the pipeline with `template` parameter.

sokol11:
Thank you @weixuanfu. Yes, I saw FeatureSetSelector but as I understood it is used to specify a static set of features as opposed to optimize the feature set composition. Am I correct in understanding that TPOT does not select features based on genetic optimization, i.e. binary coding all the features and using genetic programming to find the best feature subset? Thanks.

weixuanfu:
Your understanding is correct. TPOT cannot select features using GA without the static set of features based on _priori_ expert knowledge. Maybe it is a good enhancement function to add into TPOT.

sokol11:
Understood. Thanks!

weixuanfu:
Alternatively, you may use template like `Selector-Classifier` to indirectly select features by using GP to optimize best `Selector`.

PR #1144, opened 2020-12-16
Support arbitrary column transformers on subset of columns

edwardyu:
## What does this PR do?

Wrap `sklearn.compose.ColumnTransformer` so that other preprocessors and transformers will work on a subset of columns. This will allow for automated feature selection as well.

`tpot.builtins.ColumnTransformer` and look at changes to `base.py` to dynamically modify `config_dict`

Run classifier and regressor with `ColumnTransformer` enabled.

#1110 #1078 

## Questions:

Maybe, will let reviewer decide.

No


weixuanfu:
Sorry for the late response. I drafted a comment last week but forgot to push the green comment button. 

Could you please a demo in docs_sources for using this new transformers? Thank you for this PR.

edwardyu:
Documentation added

weixuanfu (changes requested):
Neat! I think this PR is almost ready to merge to development branch. But I think it should be an optional operator instead of putting it in all configurations under `config` since the pipeline maybe very hard to read if the input dataset has hundreds of columns (too many `include_col_` in the pipeline). Please exclude it from build-in configurations of TPOT and add a small example like we did for [FeatureSetSelector](https://epistasislab.github.io/tpot/using/#featuresetselector-in-tpot). Thank you!

edwardyu:
OK changes made. I agree, I was thinking a future feature could be making the ColumnTransformer export use the sklearn version instead of the TPOT version, since they are the same anyways and it would make the exported pipeline much more visually simpler.

weixuanfu (approved):
Thank you for this PR. I will merge it to dev branch first.

Issue #1186, opened 2021-03-07
Inclusion of more regressors and classifiers in the Model Selection


ankitrajixr:
Hi team,

Thank you for such a helpful library. While using tpot library, we found certain regressors and classifiers not included in the model selection of the machine learning pipeline.
It would be great with the addition of some regressors like Gaussian Process Regressor, Voting Regressor and classifiers like Voting Classifier, AdaBoost Classifier. 

How to recreate it?

1. User creates TPOT instance
2. User calls TPOT `fit()` function with training data

Expected result

The above-mentioned regressor and classifier are not included in the Machine learning for Model selection.

JDRomano2:
Hi @ankitrajixr, these may have previously been found to not play well with other parts of TPOT, and that may be why they are not included by default.

However, you can add any scikit-learn classifier or regressor to TPOT by simply including it in a custom configuration dictionary. Please see:
https://epistasislab.github.io/tpot/using/#customizing-tpots-operators-and-parameters

If you can use them and they perform well, we can look into adding them to the built-in configuration dictionaries. I'd recommend giving it a try and letting us know (on this thread) how they perform.

ankitrajixr:
Thank you for your response @JDRomano2 . I have tried the custom TPOT config dictionary. Below is the code snippet for it. 

```
tpot_config = {
    'kernel' : [1.0*RBF(length_scale=0.5, length_scale_bounds=(1e-05, 100000.0)),
           1.0*RationalQuadratic(length_scale=0.5, alpha=0.1),
           1.0*ExpSineSquared(length_scale=0.5, periodicity=3.0,
                                length_scale_bounds=(1e-05, 100000.0),
                                periodicity_bounds=(1.0, 10.0)),
           ConstantKernel(0.1, (0.01, 10.0))*(DotProduct(sigma_0=1.0, sigma_0_bounds=(0.1, 10.0)) ** 2),
           1.0**2*Matern(length_scale=0.5, length_scale_bounds=(1e-05, 100000.0),
                        nu=0.5)],
        'alpha': [5e-9,1e-3, 1e-2, 1e-1, 1., 10., 100.],
        'normalize_y' : [True, False],
        'optimizer' : ['fmin_l_bfgs_b']
}
```

The above code works fine for smaller datasets.

ankitrajixr:
#1191

PR #1191, opened 2021-03-17
add custom regressor classifer #1186

ankitrajixr:
#1186 

## What does this PR do?
We found certain regressors and classifiers not included in the model selection of the machine learning pipeline. Hence, we have included a few.

Contributors are:
Ankit Raj [https://github.com/ankitrajixr](url)
Darshan Chandra [https://github.com/drshnchndr](url)
Sindhu Subramanya [https://github.com/Sindhu-Subramanya](url)

### DIFF
--- a/tpot/config/classifier.py
+++ b/tpot/config/classifier.py
@@ -24,7 +24,7 @@ License along with TPOT. If not, see <http://www.gnu.org/licenses/>.
 """
 
 import numpy as np
-
+from sklearn.gaussian_process.kernels import WhiteKernel, Matern, RBF, DotProduct, RationalQuadratic, ExpSineSquared
 # Check the TPOT documentation for information on the structure of config dicts
 
 classifier_config_dict = {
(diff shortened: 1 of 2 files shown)


JDRomano2:
Hi @ankitrajixr , thanks for submitting this pull request. There are a couple of issues here:

1. The biggest one is that the tests fail. AppVeyor reports 10 errors, which look like syntax errors. Please also note that you can't put new import statements in the config dictionary files. Every tunable metaparameter must be a predefined property of the operator.
2. Once tests pass, we will need to be sure that code coverage metrics do not drop. It may be necessary to write new tests to make sure we have adequate code coverage with the test suite.
3. Before we can merge these changes we need to have some indication of their performance. This could be as simple as showing training times with and without the new operators included, and an indication of how often the new operators find their way into the final pipelines.

JDRomano2:
@ankitrajixr as another note, please also make sure to include detail describing the exact nature of the changes this pull request (when completed) will introduce. For reference, you may look at the following blog post: https://github.blog/2015-01-21-how-to-write-the-perfect-pull-request/

rachitk:
Note that this PR will fail to pass checks and the proposed regressor will not function due to its use of the kernels within the configuration dictionary, which will cause TPOT to crash when trying to evaluate any pipeline string with these kernels due to them not existing in TPOT's operator space as TPOT does not consider parameters when importing operators. 

#1218 touches on some of the changes that would need to be made to TPOT to support these kind of configurations (needing to either allow the user to pass references to objects used as parameters when instantiating TPOT or finding a way to dynamically assign those references within TPOT itself). Until those changes are made, I don't think this PR can be merged quite yet.

Issue #1246, opened 2022-04-28
Idea to make predict_proba consistent with other sklearn models/pipelines.


perib:
Currently there is not an intuitive way to check if the tpot estimator has the predict_proba function and is inconsistent with other sklearn pipelines/methods. 

## Context of the issue

The only method of figuring out if the pipeline has predict_proba is to look into the internal variable with `hasattr(tpot_instance.fitted_pipeline_, "predict_proba)`
other sklearn methods all allow you to call hasattr directly on the estimator. such as `hasattr(svc_instance, "predict_proba")`

## Process to reproduce the issue

[ordered list the process to finding and recreating the issue, example below]

1. User creates TPOT instance
2. User calls TPOT `fit()` function with training data
3. if the pipeline does not have the predict_proba method, `hasattr(tpot_instance, "predict_proba")` still returns true.

## Expected result

` hasattr(tpot_instance "predict_proba")` should return false when the pipeline does not have predict_proba, to be consistent with other sklearn methods removing the need to treat tpot as a special case.

## Current result

if the pipeline does not have the predict_proba method, `hasattr(tpot_instance "predict_proba")` still returns true.

## Possible fix

We can use this : sklearn sklearn.utils.metaestimators.available_if

I think this would be helpful for TPOTs predict_proba method.
we could simplify it to
```
@available_if(hasattr(self.fitted_pipeline_,"predict_proba")
def predict_proba(self, features):
    features = self._check_dataset(features, target=None, sample_weight=None)
    return self.fitted_pipeline_.predict_proba(features)
```

[Example with sklearn implementation of SVC](https://github.com/scikit-learn/scikit-learn/blob/4ca4f3bbcd6fb544d92886ce9690c001d8e75978/sklearn/svm/_base.py#L826)

Issue #1170, opened 2021-01-27
AttributeError: 'LinearSVC' object has no attribute 'predict_proba'


eddiebergman:
Hi,

I'm using TPOT as part of a paper benchmarking autoML tools and MCS tools (DESlib). I've run into an issue where a pipeline exposes `predict_proba` as a method but one of the pipeline steps does not actually support this and causes a crash.

## Context of the issue
TPOT has generated the following model but the `LinearSVC` step does not support `predict_proba` causing an `AttributeError: 'LinearSVC' object has no attribute 'predict_proba'` when used in further steps, i.e. `tpot_classifier.predict_proba(X_test)`. A further look at [sklearn.svm.LinearSVC](https://scikit-learn.org/stable/modules/generated/sklearn.svm.LinearSVC.html)
confirms this to be the case.

```python
tpot_classifier.fit(X, y)

if 'predict_proba' in dir(tpot_classifier): 
    tpot_classifier.predict_proba(X_test)

 AttributeError: 'LinearSVC' object has no attribute 'predict_proba'
```
This would be fine if the method was not exposed as the library DESlib checks for this but unfortunatly it is exposed without it being implemented throughout the full pipeline.

The full model is:
```python
In[.] tpot_classifier.get_params()
{'memory': Memory(location=/tmp/tmp1ft4o52c/joblib),
 'steps': [('featureunion',
   FeatureUnion(transformer_list=[('featureunion-1',
                                   FeatureUnion(transformer_list=[('functiontransformer-1',
                                                                   FunctionTransformer(func=<function copy at 0x2aaab2389d30>)),
                                                                  ('functiontransformer-2',
                                                                   FunctionTransformer(func=<function copy at 0x2aaab2389d30>))])),
                                  ('featureunion-2',
                                   FeatureUnion(transformer_list=[('functiontransformer-1',
                                                                   FunctionTransformer(func=<function copy at 0x2aaab2389d30>)),
                                                                  ('functiontransformer-2',
                                                                   FunctionTransformer(func=<function copy at 0x2aaab2389d30>))]))])),
  ('stackingestimator-1',
   StackingEstimator(estimator=LinearSVC(C=0.0001, dual=False, random_state=5,
                                         tol=1e-05))),
  ('stackingestimator-2',
   StackingEstimator(estimator=LinearSVC(C=0.001, dual=False, penalty='l1',
                                         random_state=5, tol=1e-05))),
[...]

PR #1251, opened 2022-05-19
Makes predict_proba consistent with other sklearn methods

perib:
## What does this PR do?

This pull request makes predict_proba consistent with other sklearn methods. Previously the predict_proba method was exposed when the final pipeline does not have predict_proba, that is hasattr(est, "predict_proba") returns True, but throw a runtime error when called. The new behavior is that the method is hidden when not available. Now hasattr(est, "predict_proba") returns false when predict_proba is not available in the final pipeline. Additionally the error type is changed from runetimeerror to attributeerror

Note that changed the Runtimeerror to Attribute error since that is what the sklearn VotingClassifier and SVM methods used. I'm open to suggestions if another error would be more accurate.

I added test_predict_proba_6() in tpot_tests.py in the test folder. The other test methods also test the same function. Theses were just updated to the new error type.

Also note that some unrelated tests fail due to deprecation in new versions of sklearn. Specifically sklearn.utils.testing is not sklearn.utils._testing in the more recent versions. That could be fixed in another pull request.

Other sklearn methods seem to not expose predict_proba when not available. For example, [VotingClassifier](https://github.com/scikit-learn/scikit-learn/blob/80598905e/sklearn/ensemble/_voting.py#L393) and [SVM](https://github.com/scikit-learn/scikit-learn/blob/4ca4f3bbcd6fb544d92886ce9690c001d8e75978/sklearn/svm/_base.py#L826). 

This allows users to treat tpot estimators the same as other estimators by using hasattr(est, "predict_proba") when checking if a particular model has that attribute. If this pull request is accepted, it could also be clarified in the documentation that this is an option. 

An alternative behavior would be to have predict_proba just default to predict() when probabilities are not available. However, if tpot would throw an error with predict_proba, the method should be hidden similar to the other methods in sklearn.

This was mentioned in issues #1246 , #1170

## Questions:

Yes, that would be helpful

no

### DIFF
--- a/tests/tpot_tests.py
+++ b/tests/tpot_tests.py
@@ -936,15 +936,16 @@ def test_predict_proba_2():
 
 
 def test_predict_proba_3():
-    """Assert that the TPOT predict_proba function raises a RuntimeError when no optimized pipeline exists."""
+    """Assert that the TPOT predict_proba function raises an AttributeError when no optimized pipeline exists."""
     tpot_obj = TPOTClassifier()
     tpot_obj._fit_init()
 
-    assert_raises(RuntimeError, tpot_obj.predict_proba, testing_features)
+    with assert_raises(AttributeError) as cm:
+        tpot_obj.predict_proba(testing_features)
 
 
 def test_predict_proba_4():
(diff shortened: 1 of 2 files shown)


nickotto (approved):
pulled and tested on my end with no problems!

Issue #1223, opened 2021-08-05
n_jobs Not Stopping CPU from Running at 100% on All Threads?


windowshopr:
[provide general introduction to the issue and why it is relevant to this repository]

Just performed a fresh install of TPOT (Windows 10, Python 3.7.9) using the following commands in succession (note the 'torch' now, not 'py-torch'):

```
pip install numpy scipy scikit-learn pandas joblib torch
pip install deap update_checker tqdm stopit xgboost
pip install dask[delayed] dask[dataframe] dask-ml fsspec>=0.3.3 distributed>=2.10.0
pip install tpot
```

...and I'm having an issue with Dask, specifically, even with setting `n_jobs=2`, 100% of all 12 of my cpu's threads are engaged when I start the classifier. Doesn't seem right? My dataset is quite large at 27,270 rows, and 599 columns, so I shrunk it down to 5,000 rows and kept all columns, and still it's pinnin' my CPU pretty hard. Why would this be?

![image](https://user-images.githubusercontent.com/40477419/128283399-96387cad-a23e-48ca-b8e7-706f4cf1ffaa.png)

Thanks!

windowshopr:
I'm thinking it could be because my dataset is too big, even at 5,000 rows. I shrunk it down to just a couple hundred rows and it's much quieter. Maybe once it loads all copies of the dataset into memory it would settle down?? Population size seems to play a factor as well...

perib:
when use_dask is set to True, the n_jobs argument is actually not used. This should probably be fixed. When using dask it just uses dask.compute without setting the number of threads ([here](https://github.com/EpistasisLab/tpot/blob/master/tpot/base.py#L1560)). My understanding is that this defaults to using all the threads available. 


To fix this, you need to have a dask wrapper around your .fit() command. For example:

```
import dask
from dask.distributed import Client, progress, LocalCluster

with LocalCluster(threads_per_worker=threads_per_worker, n_workers=n_workers, processes=processes,memory_limit=memory_limit) as cluster:
        with Client(cluster) as client:
            with dask.distributed.performance_report(filename=Dask_folder): #optional
                  est = tpot.TPOTClassifier()
                  est.fit(data)
```

On a single machine, I think it is best to use processes=False, n_workers=1, and threads_per_worker set to the number of cores/threads you want to use. (It should probably be benchmarked whether to have one worker and many threads or many workers with one thread, not 100% sure which option performs better).

PR #1252, opened 2022-05-20
Dask now uses njobs instead of all possible cores

perib:
## What does this PR do?

Dask now uses n_jobs instead of all possible cores. In dask.compute the parameters is now set as follows:  num_workers=self.n_jobs

Also removed the line `self.dask_graphs_ = tmp_result_scores` since this variable isn't actually used anywhere.

est = tpot.TPOTClassifier(use_dask=True, n_jobs=1)
est = tpot.TPOTClassifier(use_dask=True, n_jobs=32)

and observe the CPU usage and time to completion.

Note that within a dask client context manager, the client parameters for n_workers is used rather than the n_jobs passed into tpot.

#1223

## Questions:

Might be helpful to demonstrate an example using localcluster, and dask joblib. When within a client context manager, n_jobs in dask.compute is actually ignored in favor of the LocalCluster parameters. Since this behavior is not obvious, I think it would be helpful to include in the docs.

For example:

```
with LocalCluster(threads_per_worker=32, n_workers=1, processes=False) as cluster:
    with Client(cluster) as client:
        with dask.distributed.performance_report():

            start = time.time()
            est = tpot.TPOTClassifier(use_dask=True)
            est.fit(X,y)
            t1 = time.time() - start
            print(t1)
```

no

### DIFF
--- a/tests/tpot_tests.py
+++ b/tests/tpot_tests.py
@@ -936,15 +936,16 @@ def test_predict_proba_2():
 
 
 def test_predict_proba_3():
-    """Assert that the TPOT predict_proba function raises a RuntimeError when no optimized pipeline exists."""
+    """Assert that the TPOT predict_proba function raises an AttributeError when no optimized pipeline exists."""
     tpot_obj = TPOTClassifier()
     tpot_obj._fit_init()
 
-    assert_raises(RuntimeError, tpot_obj.predict_proba, testing_features)
+    with assert_raises(AttributeError) as cm:
+        tpot_obj.predict_proba(testing_features)
 
 
 def test_predict_proba_4():
-    """Assert that the TPOT predict_proba function raises a RuntimeError when the optimized pipeline do not have the predict_proba() function"""
+    """Assert that the TPOT predict_proba function raises an AttributeError when the optimized pipeline do not have the predict_proba() function"""
     tpot_obj = TPOTRegressor()
     tpot_obj._fit_init()
     pipeline_string = (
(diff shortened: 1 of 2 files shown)


nickotto (approved):
Had some hiccups testing the code given the mercurial nature of testing compute, but with running this test found that the workers are running properly
`from sklearn.datasets import load_breast_cancer
import tpot
import numpy as np
import dask
from dask.distributed import Client, progress, LocalCluster
import time


X,y = load_breast_cancer(return_X_y=True)

#X = np.random.rand(100,1000)
#y = np.random.randint(0,2,100)

random_state = 100
population_size= 32*5
generations = 1



with LocalCluster(threads_per_worker=1, n_workers=1, processes=False, memory_limit='64GB') as cluster:
    with Client(cluster) as client:
        with dask.distributed.performance_report():

            start = time.time()
            est = tpot.TPOTClassifier(use_dask=True, generations=generations, population_size=population_size, verbosity=2, random_state=random_state)
            est.fit(X,y)
            t1 = time.time() - start


with LocalCluster(threads_per_worker=1, n_workers=8, processes=False, memory_limit='64GB') as cluster:
    with Client(cluster) as client:
        with dask.distributed.performance_report():

            start = time.time()
            est = tpot.TPOTClassifier(use_dask=True, generations=generations, population_size=population_size, verbosity=2, random_state=random_state)
            est.fit(X,y)
            t2 = time.time() - start

print(t1)
print(t2)`

PR #1268, opened 2022-08-25
Mate operator fix

perib:
## What does this PR do?

_mate_operator has a wrapper that tries to check if all its returned values are valid pipelines. The last returned item is actually a dictionary and does not need to be evaluated. This led to the following error being thrown any time this function was called

_pre_test decorator: _mate_operator: num_test=0 'str' object has no attribute 'arity'.

I added a check in the decorator specific to _mate_operator to prevent this. Alternatively, We could also implement a generic check to skip anything that is not a pipeline/deap individual


RianKoja:
Seems nice! I get this error a lot.

PR #1280, opened 2022-12-21
Replace `np.float` with `np.float64` in `one_hot_encoder`

sarahyurick:
## Any background context you want to provide?
[NumPy 1.24.0](https://numpy.org/doc/stable/release/1.24.0-notes.html#expired-deprecations) removed the `np.float` alias, so we have to replace `np.float` with `np.float64` in `one_hot_encoder`.

Another option is to replace `np.float` with Python's `float`, although the NumPy version may be useful for consistency with NumPy arrays ([source](https://numpy.org/doc/stable/release/1.20.0-notes.html#deprecations)).

PR #1292, opened 2023-03-20
Fixing AttributeError

abdullahkavakli:
In numpy 1.24.2 (Python 3.9.16), it raises AttributeError. So, fix has implemented. If you want to get same error just run this:
```python
```
Full error is down below: 
AttributeError: module 'numpy' has no attribute 'float'. `np.float` was a deprecated alias for the builtin `float`. To avoid this error in existing code, use `float` by itself. Doing this will not modify any behavior and is safe. If you specifically wanted the numpy scalar type, use `np.float64` here. The aliases was originally deprecated in NumPy 1.20; for more details and guidance see the original release note at:
    https://numpy.org/devdocs/release/1.20.0-notes.html#deprecations

### DIFF
--- a/tpot/builtins/one_hot_encoder.py
+++ b/tpot/builtins/one_hot_encoder.py
@@ -158,7 +158,7 @@ class OneHotEncoder(BaseEstimator, TransformerMixin):
 
         Non-categorical features are always stacked to the right of the matrix.
 
-    dtype : number type, default=np.float
+    dtype : number type, default=np.float64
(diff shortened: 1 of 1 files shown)


yevIbrahimov (approved):

yevIbrahimov:
@jay-m-dev please approve, thank you!

perib:
This is a duplicate of #1280  which has just been merged.

Issue #1327, opened 2023-10-11
In python 3.12, Get error after importing module.


115dkk:
Compile error

https://github.com/EpistasisLab/tpot/issues/791

As you can see, there was a warning that 'imp' is deprecated. However, since then, there hasn't been much improvement...

In Python 3.12, that feature was completely removed, and now I get compile errors because of it.

## Process to reproduce the issue

Import TPOT by any means and run your code in python 3.12.

## Expected result

![image](https://github.com/EpistasisLab/tpot/assets/33408534/546bd790-527b-4e90-984a-b8ec5a26757b)

## Current result

ModuleNotFoundError: No module named 'imp'

## Possible fix

I fixed this by manually erasing "import imp" in base.py of tpot.

or.... Perhaps https://pypi.org/project/zombie-imp/ could be used...

jamesmyatt:
Is this fixed in [v0.12.2](https://github.com/EpistasisLab/tpot/releases/tag/v0.12.2)? Looks like https://github.com/EpistasisLab/tpot/pull/1331 is part of that release.

PR #1331, opened 2023-11-28
remove deprecated imp, fix docstring warning

perib:
imp is deprecated in Python 12.  This was only used in one function which I replaced with the equivalent function found in the types module.

Also, the docstrings that contained backslashes were causing a warning to be called every time that was imported. These were probably intended for latex formatting, but Python reads them differently. I turned those into literal docstrings so that the Python interpretation of escape sequences was ignored, which stopped the warnings.

### DIFF
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -28,7 +28,7 @@ import random
 import inspect
 import warnings
 import sys
-import imp
+
 from functools import partial
 from datetime import datetime
 from multiprocessing import cpu_count
@@ -38,7 +38,7 @@ import errno
 
 from tempfile import mkdtemp
 from shutil import rmtree
-
+import types
 import numpy as np
 from pandas import DataFrame
 from scipy import sparse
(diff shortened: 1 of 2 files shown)


perib:
Forgot to mention this issue was brought up in #1327

jay-m-dev:
The suggested fix seems to work, but the [Python docs](https://docs.python.org/3.11/library/imp.html) suggest using importlib instead of imp.
We should further assess which solution is better: the suggested `types.ModuleType` or `importlib.util.module_from_spec()` as documented [here](https://docs.python.org/3/library/importlib.html#importlib.util.module_from_spec).

tpot/base.py
jay-m-dev, line 1: The [Python docs](https://docs.python.org/3.11/library/imp.html) recommend using importlib instead of the imp package. However, upon further research and testing types.ModuleType works here as well. There's no clear benefit of using one over the other.
Accepting this update as-is.

jay-m-dev (approved):
The [Python docs](https://docs.python.org/3.11/library/imp.html) recommend using importlib instead of the imp package. However, upon further research and testing types.ModuleType works here as well. There's no clear benefit of using one over the other.
Accepting this update as-is.

PR #1335, opened 2023-12-08
Update requirements

gatl:
Due to the use of the decorator 
`sklearn.utils.metaestimators.available_if()` by TPOT, declare as the minimum version required of scikit-learn the version that introduced this decorator, that was 1.0.

## What does this PR do?
Update the requirements for TPOT.

This is just a value change.

Verify that TPOT no longer works with scikit-learn version 0.22.0 or any other version prior to 1.0, but it works with version 1.0 and above.

I think none is needed.

There are none, other than TPOT failing to work when all its stated requirements are satisfied.

None is provided.

## Questions:

I think there is one reference that would benefit of attention in the installation section, despite referring to dask.

No, it does not.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 deap>=1.2
 nose==1.3.7
 numpy>=1.16.3
-scikit-learn>=0.22.0
+scikit-learn>=1.0
 imbalanced-learn>=0.7.0
 scipy>=1.3.1
 tqdm>=4.36.1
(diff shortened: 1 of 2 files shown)


perib:
The setup.py file should also be updated to match. If you can make that change, I can merge this.
Thanks for the contribution!

gatl:
Oh, sorry. Right.
But now I see you did it already. Thanks.

PR #1338, opened 2023-12-12
Improve error message

gatl:
Unfortunately there is no fine granularity when an error occurs. That makes it more relevant to list as many possibilities for the causes as possible. In this patch, one additional cause is added to the ones already presented (no optimization done due to timeout from max_eval_time_mins).

## What does this PR do?
Presents one additional cause to the exception that the user may get. It took me some time to understand the error and this may save other people from this work.

Review the change to the message.

No testing needed, just validate the change to the test.

It took me many hours of work to figure out why this `RuntimeError` was being raised. Hopefully, others in the same situation can diagnose their situation in less time.

No relevant issues that I am aware of.

It is not.

## Questions:

No, there is no need to do so. However, there is no mention of what happens when the timeouts are reached. The present `RuntimeError` may not be the full story, so I will leave this chore to someone with greater familiarity with the code.

No additional dependency is added.

### DIFF
--- a/tpot/base.py
+++ b/tpot/base.py
@@ -935,7 +935,8 @@ class TPOTBase(BaseEstimator):
                 raise RuntimeError(
                     "There was an error in the TPOT optimization "
                     "process. This could be because the data was "
-                    "not formatted properly, or because data for "
+                    "not formatted properly, because a timeout "
+                    "was reached or because data for "
                     "a regression problem was provided to the "
                     "TPOTClassifier object. Please make sure you "
                     "passed the data to TPOT correctly. If you "


jay-m-dev (approved):
I was able to trigger this RuntimeError with:
`from tpot import TPOTClassifier
from sklearn.datasets import load_breast_cancer
import numpy as np

breast_cancer = load_breast_cancer()
X_train, X_test, y_train, y_test = train_test_split(breast_cancer.data.astype(np.float64),
    breast_cancer.target.astype(np.float64), train_size=0.75, test_size=0.25, random_state=42)

est = TPOTClassifier(generations=5, population_size=50, verbosity=2, random_state=42, max_eval_time_mins=0.0001)
est.fit(X_train, y_train)`

But also had to update the `timeout` var set in base.py.

Issue #1341, opened 2024-01-27
TPOT report error


sch401:
Today when I ran a machine learning project using TPOT, it reported the error below:
```
D:\rebirth - default\venv_test\Lib\site-packages\tpot\builtins\__init__.py:36: UserWarning: Warning: optional dependency `torch` is not available. 
- skipping import of NN models.
  warnings.warn("Warning: optional dependency `torch` is not available. - skipping import of NN models.")

Generation 1 - Current best internal CV score: -inf
Traceback (most recent call last):
  File "D:\rebirth - default\venv_test\Lib\site-packages\tpot\base.py", line 817, in fit
    self._pop, _ = eaMuPlusLambda(
                   ^^^^^^^^^^^^^^^
  File "D:\rebirth - default\venv_test\Lib\site-packages\tpot\gp_deap.py", line 285, in eaMuPlusLambda
    per_generation_function(gen)
  File "D:\rebirth - default\venv_test\Lib\site-packages\tpot\base.py", line 1183, in _check_periodic_pipeline
    self._update_top_pipeline()
  File "D:\rebirth - default\venv_test\Lib\site-packages\tpot\base.py", line 935, in _update_top_pipeline
    raise RuntimeError(
RuntimeError: There was an error in the TPOT optimization process. This could be because the data was not formatted properly, or because data for a regression problem was provided to the TPOTClassifier object. Please make sure you passed the data to TPOT correctly. If you enabled PyTorch estimators, please check the data requirements in the online documentation: https://epistasislab.github.io/tpot/using/

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "d:\rebirth - default\tpot优化.py", line 169, in <module>
    tpot.fit(training_features, training_target)
  File "D:\rebirth - default\venv_test\Lib\site-packages\tpot\base.py", line 864, in fit
    raise e
  File "D:\rebirth - default\venv_test\Lib\site-packages\tpot\base.py", line 855, in fit
    self._update_top_pipeline()
  File "D:\rebirth - default\venv_test\Lib\site-packages\tpot\base.py", line 935, in _update_top_pipeline
    raise RuntimeError(
RuntimeError: There was an error in the TPOT optimization process. This could be because the data was not formatted properly, or because data for a regression problem was provided to the TPOTClassifier object. Please make sure you passed the data to TPOT correctly. If you enabled PyTorch estimators, please check the data requirements in the online documentation: https://epistasislab.github.io/tpot/using/
```
[...]

PR #1343, opened 2024-02-20
update compatibility with scikitlearn 1.4

perib:

## What does this PR do?

This resolves the error encountered in #1341 . The latest version of sklearn updated the internal function  `def _fit_and_score` to take in an additional score_params parameter. When TPOT attempted to evaluate a pipeline with this function, it would throw an error because this parameter was missing. 

The parameter was introduced in this PR https://github.com/scikit-learn/scikit-learn/pull/26896/files

I tested with the following code. Runs with the fix but not without the fix.
```
import numpy as np

iris = load_iris()
    iris.target.astype(np.float64), train_size=0.75, test_size=0.25, random_state=42)

est = TPOTClassifier(generations=5, population_size=50, verbosity=2, random_state=42)
est.fit(X_train, y_train) 
```

## Questions:

Updates the dependencies of scikit learn to version 1.4 . This change is incompatible with previous versions of scikit-learn

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,7 +1,7 @@
 deap>=1.2
 nose==1.3.7
 numpy>=1.16.3
-scikit-learn>=1.1.3
+scikit-learn>=1.4.1
 imbalanced-learn>=0.7.0
 scipy>=1.3.1
 tqdm>=4.36.1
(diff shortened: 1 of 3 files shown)


jay-m-dev (approved):
Tested on Python 3.10.12 and 3.12.1

Issue #1366, opened 2025-04-04
Question Regarding Genetic Feature Selection


boguskiP:
Is there a simple/built-in method of ensuring that if you extract 2+ subsets from the same dataset (using Genetic Feature Selection), that the features found in the first subset are not found in another subset? I.e., Obtain a series of series of subsets that are "symmetrically different" mathematically. 

I'm wondering since I want to have each subset have it's own 'branch' before coming all together and the presence of duplicated features makes understanding the 'feature importance' more difficult. Just asking to ensure I'm not coding for something that exists but I couldn't find it.

perib:
There are two options for doing evolutionary feature selection in TPOT (which you can learn more about in Tutorial 4: https://github.com/EpistasisLab/tpot/blob/main/Tutorial/4_Genetic_Feature_Selection.ipynb)
Each FSSNode node picks one predefined subset of features. And GeneticFeatureSelectorNode gets a list of features and dynamically learns any subset of those features per node. A single GeneticFeatureSelectorNode will never select the same feature twice.

So for instance FSSNode you pass in
"subset a": [ 1,2,3],
"subset b": [ 5,6,7],
"subset c": [ 8,9,4],

And it will pick one of the three sets.

For GeneticFeatureSelectorNode, you pass in all features ([0,1,2,3,4,5,6,7,8,9]) and it will pick any subset. for example [1] or [3, 5, 7,8,9] or [6, 8]. 

Where you position these in your pipeline determines whether or not repeat features are selected. Here are a few examples using the FFSNode, but the same idea applied to the GeneticFeatureSelectorNode.

With the FSSNode, each node will select one of the predefined subsets of features. When using graph pipelines, TPOT will never actually create two duplicates of the same feature in a given pipeline. If crossover or mutation yields a pipeline with say 2 feature set B nodes in the leaves, those will actually be merged into one node with two (or more) outputs to their corresponding nodes. So repeat features wouldn't occur. (This may not work exactly with the GeneticFeatureSelectorNode since there is much more diversity in possible features and they have to exactly match to be merged. When using graph pipelines, you should not use GeneticFeatureSelectorNode  in the leaves. Instead you should either use FFSNore in the leaves or use a sequential pipeline of a GeneticFeatureSelectorNode  followed by a graphpipeline)

[...]

PR #1368, opened 2025-04-09
fix fss node not returning T/F and not handling cases with only one f…

perib:
## What does this PR do?

FSSNode was not returning True or False for the mutation/crossover operations as required. 
Additionally, when only a single feature set was passed in, there would be an error thrown. This was found in issue #1366 

```
subset_1_dict  = { "sense_strands" :  ['a', 'b', 'c']}
fss_1 = FSSNode(subsets=subset_1_dict)

ind1 = fss_1.generate()

for i in range(1000):
    ind1.mutate()
```

Before this would throw an error like : 

`File "/scratch2/users/mpb703/conda_environments/tpot_v1/lib/python3.10/site-packages/tpot/search_spaces/nodes/fss_node.py", line 113, in mutate self.selected_subset_name = rng.choice(names) File "numpy/random/_generator.pyx", line 803, in numpy.random._generator.Generator.choice ValueError: a cannot be empty unless no samples are take `

### DIFF
--- a/tpot/search_spaces/nodes/fss_node.py
+++ b/tpot/search_spaces/nodes/fss_node.py
@@ -109,14 +109,21 @@ class FSSIndividual(SklearnIndividual):
     def mutate(self, rng=None):
         rng = np.random.default_rng(rng)
         #get list of names not including the current one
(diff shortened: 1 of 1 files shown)


perib:
bump

Issue #1379, opened 2025-06-18
Modernize PyPI deployment


jay-m-dev:
Modernize PyPI deployment to use pyproject.toml instead of the legacy setup.py build process.

jay-m-dev:
It would be a great idea to add a test for the built pip package before pushing it to PyPI.
Description of issue:
- A build configuration can be created such that the folder structure in the pip package could be non-standard and cause import errors.

example pyproject.toml configuration to look into:
[tools.setuptools]
package-dir = {"" = "src"}
packages = ["<packagename>"] // may also need to test if this issue arises specifically due to a mismatch in the package name and folder name (e.g. heros/ vs skheros)

[tool.setuptools.packages.find]
where = ["src"]
#include = ["heros*"] // test if it matters that this is the folder name and not the package name (e.g. skheros instead?)
#exclude = ["tests*"]

The main thing is to ensure package imports work **before** pushing to PyPI

PR #1381, opened 2025-07-25
Update dependency management

chimaerase:

## What does this PR do?

Provide baseline updates to use `pyproject.toml` + `poetry` for more automated dependency management and deployment (#1379), as well as differentiating between CPU- and GPU-targeted installs.  Likely some minor deployment-related details are still missing here (which I can't test), but I'm hoping this is a useful baseline to work from.  Builds on an earlier [draft PR I submitted to the tpot2 repo](https://github.com/EpistasisLab/tpot2/pull/170), but now CI tests are updated / passing.

Should resolves several issues:

1.  The need to hard-code dependency versions for sub-packages
    I left TODO's in `pyproject.toml` with a partial list of those I didn't need to make the tests pass.
3.  Auto-inclusion of GPU (CUDA) support that adds unnecessary bloat when not in use
4.  Needing multiple files to track dev/production dependencies separately 
5. Makes `update_checker` optional (opt-in)
    In context of my closed source apps that use TPOT, this check is overkill that adds unnecessary / confusing output that end users aren't familiar with & don't know how to resolve.  I see the value for other users, but there doesn't appear to be any way, e.g. with configuration, to silence the very-aggressive update checks (can appear minutes after a new TPOT release).

New local dev install process would be, e.g.:

```
# Install with xghoost-cpu (no GPU support)
poetry install --extras cpu
# Install with xghoost (includes NVIDIA CUDA support -- only needed when available)
poetry install --extras cuda
# Install development packages
poetry install --with dev
```

Downstream pypi downloads of published versions should support extras, e.g. `pip install tpot[cpu]` or similar.

`pyproject.toml`

Tests in tox.ini are already passing.  There may be some additional testing needed to make this work in deployment to pypi.  I drafted the code for that, but don't have access to test deployments for TPOT (or any experience publishing packages to pypi yet).

I'm again short on time to help with this, so may not be able to address comments quickly (or at all). I'm hoping this is a big boost in a helpful direction, and happy for others to take the ball and run with it.

[you can link directly to issues by entering # then the number of the issue]

## Questions:

### DIFF
--- a/.github/workflows/publish_package.yml
+++ b/.github/workflows/publish_package.yml
@@ -20,14 +20,12 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install build twine
+        curl -sSL https://install.python-poetry.org | python3 -
 
     - name: Build package
-      run: python -m build
+      run: poetry build
 
     - name: Upload to PyPI
       env:
-        TWINE_USERNAME: __token__
-        TWINE_PASSWORD: ${{ secrets.PYPI_API_TOKEN }}
-      run: twine upload dist/*
-  
+        POETRY_PYPI_TOKEN_POETRY: ${{ secrets.PYPI_API_TOKEN }}
+      run: poetry publish
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -21,6 +21,10 @@ jobs:
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
-        pip install tox tox-gh-actions
+        sudo apt-get update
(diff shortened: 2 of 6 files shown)


chimaerase:
Changed this to "ready for review".  Likely there are a few minor issues remaining, especially as they relate to pushing to PyPi, but worth looking over at this point with tests passing.  Not sure I can commit to polishing this, but I hope it's at least a very significant boost toward the final solution.

chimaerase:
Attempting updates to this PR to resolve with conflicting updates applied to `main` after submission, and to use `uv`, which seems to recently be the de-facto Python packaging tool.

PR #1391, opened 2026-03-06
Remove `stopit` imports, references, and dependency

reevesc7:
## What does this PR do?

This PR removes `stopit` as a dependency and all (non-archived) references to it.

Removed code references include `tpot.utils.eval_utils.optimize_objective()` and `tpot.utils.eval_utils.parallel_optimize_objective()`.

- Altered code: `tpot/utils/eval_utils.py`
- Altered dependencies: `pyproject.toml`
- Altered reported dependencies: `README.md`

A fresh install and standard tests may be appropriate to validate that the dependencies resolve properly.

The removed references to `stopit` were not referenced anywhere else within TPOT, but their removal would cause incompatibilities with any external scripts calling `tpot.utils.eval_utils.optimize_objective()` or `tpot.utils.eval_utils.parallel_optimize_objective()`.
...Maybe a deprecation warning would be appropriate, but as I describe below, there was already a deprecation warning being printed related to these functions.

`stopit` imports `pkg_resources`, which was deprecated and is now removed.
For some time, `pkg_resources` printed a deprecation warning stating that the module would be removed in the near future.
That future has come, and now fresh installations may no longer function due to `stopit` still attempting to import `pkg_resources`.
In fact, I discovered that `pkg_resources` has been removed when freshly installing TPOT to make a different pull request.

Unfortunately, `stopit` is no longer maintained.
The last update to its repository occurred 3 years ago.
The fix for the `pkg_resources` issue is simple, and the `stopit` repository has the fix waiting in a pull request.
However, the request has remained open for a year and a half.

Given that `stopit` is unlikely to regain functionality in the near future and that the TPOT internals do not utilize the module, I figured it would be best to remove the dependency.

Alternatively, [spraakbanken/stopit2](https://github.com/spraakbanken/stopit2) is a fork of `stopit` which implements the `pkg_resources` fix.
If maintainers desire to keep `stopit`-like functionality within TPOT, `stopit2` may be a fitting choice.

None that I found during a quick search.

EDIT: PR #1394 is an alternate solution, and *either* this PR or that one should be merged.

## Questions:

- Do the docs need to be updated? *No, they do not mention `stopit`*
- Does this PR add new (Python) dependencies? *No, but it removes a dependency*


reevesc7:
I have changed this PR's base branch to `EpistasisLab:dev`.

reevesc7:
As another alternative solution to removing `stopit` dependency, TPOT may add a `setuptools>=70,<82` dependency.

`pkg_resources` still imports in `setuptools` through v81.0.0, and the warning about the imminent removal of `pkg_resources` prints in that version.

This is not a nice solution, as `stopit` is the only code that directly depends on `setuptools`.
However, as `stopit` expects `setuptools` to be bundled with Python distributions (as is no longer the case) and is unmaintained, TPOT *could* shoulder the dependency itself.

---

Additionally, I have opened PR #1394, as an additional alternate solution which tricks `stopit` into using a fake `pkg_resources`.
Only one of these two PRs should be merged.

